PR #237, opened 2014-12-29
Feature/sale distribution consumable

mbayopanda:
- In this PL we add the column `is_distributable` in the table `sale` in bhima.sql
- We list only distributions which are distributable (`si_distributable=1`) in `Stock_managment/Stock distribution` 
- We fixed the button `view sale details` in `Stock_managment/Stock distribution`


client/src/partials/stock/exit/distribution.js
```diff
@@ -67,9 +67,8 @@ angular.module('bhima.controllers')
 
     function startup(model) {
       angular.extend($scope, model);
-
+      $scope.ledger.data = $scope.ledger.data.filter(function (data) { return data.is_distributable[0] === 1; });
```
mbayopanda: The *function debtor* in **ledger.js** is used in more than one place, if i change the query clause 'where' it's will affect other place in the application.

jniles: Aha.  Okay, not a problem.

PR #246, opened 2015-01-03
Added ability to edit patient details

jmcameron:
The patient edit form allows the user to select a patient (using the 'find-patient' directive) and then switches to editng the selected patient.

All editing functionality is impemented except for swapping the debitor group.

This update includes an update to the 'location-select' directive that supports an 'initial-location' attribute.  Using the existing 'defaultVillage' attribute did not seem to work right for initial location values so I added the new option for 'initial-location' attribute.  This should be backwards compatible, although it may be possible to merge the functionality and only have one.


client/src/partials/patient_edit/edit_patient.html
```diff
@@ -0,0 +1,227 @@
...
+      <div class="col-xs-5">
+      </div> <!-- End column 5 row -->
+    </div> <!-- end row -->
+  </div>
+</main>
+
+<footer>
+  <div class="pull-right">
+    <a id="submitPatient" ng-click="updatePatient()"
+       ng-disabled="session.failedSessionValidation || location.$invalid || details.$invalid"
```
jniles: Is it possible to take the `$invalid` logic like this out of the template?  Maybe include them as part of the session.failedSessionValidation check?

client/src/partials/patient_edit/edit_patient.js
```diff
@@ -0,0 +1,269 @@
...
+  '$routeParams',  
+  '$translate',
+  'connect',
+  'validate',
+  'messenger',
+  'util',
+  'appstate',
+  function ($scope, $routeParams, $translate, connect, validate, messenger, util, appstate) {
+    var dependencies = {},
+        original_patient_data = null,
```
jniles: For JS variables, we prefer to us mixedCase.  Unfortunately, the database is constructed using underlines_not_spaces syntax, so we cannot get away from the underlines in some cases.  Could you change `original_patient_data` and `editable_fields`, etc, to mixedCase?

client/src/partials/patient_edit/edit_patient.js
```diff
@@ -0,0 +1,269 @@
...
+    $scope.initialiseEditing = function initialiseEditing (selectedPatient) {
+      if (selectedPatient && 'uuid' in selectedPatient && selectedPatient.uuid) {
+        patient_uuid = $scope.patient_uuid = selectedPatient.uuid;
+	dependencies.patient.query.where[0] = 'patient.uuid=' + patient_uuid;
+	validate.process(dependencies)
+	  .then(startup);
+	session.mode = 'edit';
+      }
+      else {
+	messenger.danger($translate('ERR_MISSING'));
```
jniles: I cannot find a translation for this?  Could you add a namespaced translation in en.json?

client/src/partials/patient_edit/edit_patient.html
jmcameron: I do not know how to access the 'location' form from within the 'patientEdit' controller.  I'll fix the other things and then get with you to do this later as a separate pull request.  I have a few other cleanup things that I can do at the same time.

client/src/partials/patient_edit/edit_patient.js
jmcameron: I did not realize that 'ERR_MISSING' was not translated.  I copied it from 'swap_debtor'.
I will add a translation for this, but I'm having second thoughts on exposing this internal and hopefully-rare error to end users.   If we decide not to expose this, I'll add it to my next round of improvements on the form (in another pull request).

client/src/partials/patient_edit/edit_patient.html
jniles: Okay, sounds good.

client/src/partials/patient_edit/edit_patient.js
jniles: Sounds like a discussion we should all have as a team.  We can add that discussion to the agenda.

PR #253, opened 2015-01-05
Fix translation fr

lomamech:
Fixing translation error on fr.json


client/src/i18n/fr.json
```diff
@@ -511,168 +511,6 @@
     "ALL_PROJECTS" : "Tous Projets"
   },
 
-  "PATIENT_STANDING"     : {
```
jniles: Why was this all deleted?

PR #256, opened 2015-01-05
Feature/edit patient cleanup

jmcameron:
Implemented several updates to the edit patient form, including modifying/adding patient database columns.  See the commmits for details.


client/src/partials/patient_edit/edit_patient.html
```diff
@@ -41,7 +41,7 @@
                   <div class="form-group" >
                     <label for="patient-dob" class="col-xs-3 control-label">{{ "PATIENT_REG.DOB" | translate }}</label>
                     <div class="col-xs-9">
-                      <input required id="patient-dob" type="date" class="form-bhima" ng-model="patient.dob" min="1900-01-01" max="{{ getMaxDate() }}">
+                      <input required id="patient-dob" type="date" class="form-bhima" ng-model="patient.dob" min="{{ getMinDate() }}" max="{{ getMaxDate() }}">
```
jniles: Would it be possible to cache both of these in a scope variable, rather than in a function?  Angular will call `getMinDate()` and `getMaxDate()` (which doesn't change after module load) repeatedly when redrawing the view.  Something like `session.minHtmlDate` and `session.maxHtmlDate` could be calculated and exposed to the view on module load.  I'm pretty sure this is a lighter weight solution.

client/src/partials/patient_edit/edit_patient.js
```diff
@@ -96,38 +99,43 @@ angular.module('bhima.controllers')
...
+      $scope.patient.current_location_id = uuid;
     };
 
+
+    // Validate the dates and locations
     function customValidation() {
-      var dates = validateDates();
-      var locations = validateLocations();
+      var badDOB = invalidDOB();
+      var badLocations = invalidLocations();
```
jniles: This is much more semantic.  Awesome!

client/src/partials/patient_edit/edit_patient.html
jmcameron: Ok, I updated the pull request as requested.  I just put them in $scope.  Let me know if you prefer them in session.

PR #265, opened 2015-01-06
Debitor Group Update

lomamech:
Changing the structure of table debitor_group, Updated of reauest for  inclusion of the column lock for table debitor_group and prevent billing for a patient belonging to a locked debitor_group


server/models/bhima.sql
```diff
@@ -498,12 +498,10 @@ create table `debitor_group` (
   `name`                varchar(100) not null,
   `account_id`          int unsigned not null,
   `location_id`         char(36) not null,
-  `payment_id`          tinyint unsigned not null default '3',
```
jniles: `payment_id` references the table `payment`.  You can remove the table `payment` as well, since it is no longer being used.

jniles: Also, since you have made several SQL changes, can you provide update files, like @jmcameron to help ease database migration?

jniles:
@lomamech , the files `server/models/updates/2015-01-05-mod-03.sql` and `server/models/updates/2015-01-05-mod-02.sql` were deleted by @jmcameron in a previous pull request, and you have re-added them.  Please delete them in a new commit and I can merge this pull request.

jniles:
This looks much better, thanks.

PR #266, opened 2015-01-06
Update Fiscal Year Creation

jniles:
In this PR:
  - Fixed bugs:
     - Correct number of periods displayed
     - Dates normalized between the client and server
  - Added edit fiscal year functionality
  - The client sends a post request with all data at once, rather than a staggered fetch/post for opening balances
  - Only set opening balances for the first year created
  - User now sets the parent fiscal year
  - Removed confusing "view opening balances" functionality
  - Modularized code into well-commented controllers.


DedrickEnc:
Thank you Jon for the fiscal year management pull request, here are some feedback : 
- a parent fiscal year must have only one child fiscal year, so that there will not be some conflict and confusion between fiscal year.
- a opening balance operation must write to the general ledger, so that we will be able to provide correct report.

PR #269, opened 2015-01-07
Remove hardcoded minimum birth year in patient register/edit forms

jmcameron:
Added a global value for the minimum patient DOB in the util service:    util.minimumPatientDate.

Updated DOB year-out-of-range error message to include tokens for <min> and <max> which are replaced in the form controller code.

Updated patient register/edit forms to use util.minimumPatientDate and eliminate any hardcoded dates in the form html and controller code.

### DIFF
--- a/client/src/partials/patient_registration/patient.js
+++ b/client/src/partials/patient_registration/patient.js
@@ -12,12 +12,14 @@ angular.module('bhima.controllers')
   'uuid',
   function ($scope, $q, $location, $translate, connect, messenger, validate, appstate, util, uuid) {
 
-    var dependencies = {};
-    var defaultBirthMonth = '06-01';
+    var dependencies = {},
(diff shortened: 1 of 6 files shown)


client/src/partials/patient_registration/patient.js
```diff
@@ -131,18 +140,24 @@ angular.module('bhima.controllers')
...
       }
 
-      if (extractYear) {
+      if (yearOfBirth) {
+
+	// NOTE: The following checks on the yearOfBirth are never executed with
+	//       the html5+angular date input field since the form value becomes
+	//       undefined when invalid and is caught by the previous check.
+	//       Leaving them in as a precaution in case the input form behavior
+	//       changes in the future.
```
jniles: This is a really clean message.  Thanks for taking the investment time to produce quality code.

PR #270, opened 2015-01-07
Feature/stock loss management

mbayopanda:
In this PR : 
- we get the unit price of a lot when we declare a loss
- we have a report of loss (loss_records)
- Grouping loss data by document_uuid (we have simply an array which contain correct data, the ui is still poor... (loss_record) :-) 
- fixing stock management links for reports


client/src/partials/stock/loss/loss.js
```diff
@@ -280,6 +280,24 @@ angular.module('bhima.controllers')
...
+          tables : {
+            'stock' : { columns : ['purchase_order_uuid']},
+            'purchase_item' : { columns : ['quantity', 'unit_price']}
+          },
+          join : ['stock.inventory_uuid=purchase_item.inventory_uuid'],
+          where : ['stock.tracking_number=' + item.tracking_number]
+        }
+      };
+
+      validate.process(dependencies, ['itemPrice'])
```
jniles: If this is called multiple times, you will want to use `validate.refresh(dependencies, array)`;

server/models/updates/2015-01-05-mod-02.sql
```diff
@@ -0,0 +1,14 @@
+-- Updates to patient data
```
jniles: @mbayopanda , these were removed in a separate pull request.  Please remove them in a separate commit.

client/src/partials/stock/loss_record/loss_record.js
```diff
@@ -0,0 +1,208 @@
...
+.controller('stock.loss_record', [
+  '$scope',
+  '$timeout',
+  '$routeParams',
+  'util',
+  'validate',
+  'exchange',
+  function ($scope, $timeout, $routeParams, util, validate, exchange) {
+    // TODO add search (filter)
+    // TODO add sortable (clickable) columns
```
jniles: TODOs are excellent for keeping track of progress.  It's good to see you using them.

client/src/partials/stock/loss_record/loss_record.js
```diff
@@ -0,0 +1,208 @@
...
+    function totalLoss() {
+      return $scope.model.loss.data.length;
+    }
+
+    function groupingLoss(data){
+      // Grouping loss data by document_uuid
+      // In the case where we have more than one item in a loss document
+    	var loss_array = [];
+
+    	for(var i in data){
```
jniles: To make this more semantic, you should use `Array.forEach(fn);`.  Your code will be much more readable.  For example, here:
```javascript
data.forEach(function (item) {
  var temp = item.document_uuid;
  newLine.push(item);
  // etc...
});
```
Please also use mixedCase instead of under_scores.

client/src/partials/stock/stock.js
```diff
@@ -172,7 +179,8 @@ angular.module('bhima.controllers')
       }*/
 
       var path = config.modules.indexOf(defn) > -1 ? defn.link + '/' + $scope.depot.uuid
-        : (config.utilities.indexOf(defn) > -1 ) ? defn.link+ '/' + $scope.depot.uuid : defn.link;
+        : (config.utilities.indexOf(defn) > -1 ) ? defn.link+ '/' + $scope.depot.uuid 
+        : (config.reports.indexOf(defn) > -1 ) ? defn.link+ '/' + $scope.depot.uuid : defn.link;
```
jniles: This is very hard to read.  What is happening here?  If it is possible to write out these routes explicitly, it would be easier for your fellow developers.

server/models/updates/2015-01-05-mod-03.sql
```diff
@@ -0,0 +1,10 @@
+-- Updates to patient data
```
jniles: This was removed as well.  See #262

jniles:
@mbayopanda As soon as `server/models/updates/2015-01-05-mod-03.sql` and `server/models/updates/2015-01-05-mod-02.sql` are removed, I will merge this PR.

mbayopanda:
I have just update this pull request, someone can see it and merge it if it's possible

PR #271, opened 2015-01-07
Paiement for convention

lomamech:
Fixed problem rounding of Pay box for the payment of conventions


client/src/partials/primary_cash/income/convention/convention.html
```diff
@@ -48,11 +48,11 @@
             <tr ng-repeat="overview in overviews">
               <td>{{ overview.reference }}</td>
               <td>{{ overview.description }}</td>
-              <td align="right">{{ overview.balance | number:4 | currency }}</td>
+              <td align="right" ng-bind-html="overview.balance | intlcurrency:model.enterprise.data[0].currency_id"></td>
             </tr>
             <tr>
               <td>{{ 'CONVENTION.TOT' | translate}}</td>
-              <td colspan="2" align="right">{{ som | number:4 | currency }}</td>
+              <td colspan="2" align="right" ng-bind-html="som | intlcurrency:model.enterprise.data[0].currency_id"></td>
```
jniles: The expression `intlcurrency:model.enterprise.data[0].currency_id` is not associated to a particular enterprise.  If there were two enterprises, this code would break the application.  Instead of loading all enterprises using `dependencies`, you should get the current registered enterprise from appstate.  Such as this:
```javascript
appstate.register('enterprise', function (enterprise) { $scope.enterprise = enterprise; });
```

Then in your html, you can use
```html
<td colspan="2" align="right" ng-bind-html="som | intlcurrency:enterprise.currency_id"></td>
```

client/src/partials/primary_cash/income/convention/convention.js
```diff
@@ -85,8 +86,8 @@ angular.module('bhima.controllers')
       query : {
```
jniles: With the suggestion above, you can remove this dependency on `enterprise`.

jniles:
Looks good to me.  Thanks.

PR #276, opened 2015-01-07
Renamed SQL update file to include date in filename

jmcameron:


jniles:
@jmcameron Whoops, thanks for this.

PR #277, opened 2015-01-07
Corrections to Fiscal Year creation module

jniles:
In this PR:
 - Clicking on "Create a new fiscal year" whilst creating a fiscal year refreshes the page instead of doing nothing.
 - Removed all code pertaining to closing accounts as this was incorrect (fiscal years can be closed into multiple accounts).  The selection is now replaced with a warning, telling the use to close any open previous fiscal years.
 - Editing a fiscal year is now limited to simply the name.  Everything else is readonly.


jniles:
@DedrickEnc, could you please take a look at this and make suggestions?

There is an additional bug in which the debits and credits to opening balances do not need to sum to zero, but I would prefer to fix that in a separate pull request.

PR #308, opened 2015-01-11
Added README document for server/models folder

jniles:
At the request of @jmcameron, I've made a document describing the files in `server/models`, and their purpose.


server/models/README.md
```diff
@@ -0,0 +1,55 @@
+Bhima Models
+============
+
+This document describes the contents of the `/models` directory.
+
+The directory contents are listed below:
+```bash
```
jmcameron: What is 'bash' doing here?

jmcameron:
This README is very useful.  Thanks!

server/models/README.md
jniles: It renders it nice and pretty in github-flavored markdown.  I guess it's probably going to be read through a text editor, so I'll take it out.

PR #309, opened 2015-01-11
Enable edit patient restart

jmcameron:
Currently, when editing a patient, once you start editing a patient, you cannot switch to another patient without going to some other page. 

This PR adds a button on the right side of the Patient Details edting area (only when editing).  The user can click on the button to switch back to the search mode to find another patient to edit.

Currently when the user is editing a patient and clicks on the  main menu 'Edit Patient Details' item, nothing happens.  I'm sure it would be possible to fix this, but having an button for that purpose may be simpler for users.

Also disabled the automatic logging of the exchange rate on ever startup.

### DIFF
--- a/client/src/partials/patient_edit/edit_patient.js
+++ b/client/src/partials/patient_edit/edit_patient.js
@@ -189,9 +189,23 @@ angular.module('bhima.controllers')
       customValidation();
     });
 
+    // Restart the search
+    $scope.restartSearch = function () {
+      originalPatientData = null;
+      $scope.patient = { origin_location_id: null, current_location_id: null };
+      $scope.initialOriginLocation = null;
+      $scope.initialCurrentLocation = null;
+
+      // Make sure the patient query gets processed
+      if ('processed' in dependencies.patient) {
+	dependencies.patient.processed = false;
+      }
+
+      session.mode = 'search';
+    };
 
     // Define the function that switches to the edit mode
-    $scope.initialiseEditing = function initialiseEditing (selectedPatient) {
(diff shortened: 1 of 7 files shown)


client/src/partials/patient_edit/edit_patient.js
```diff
@@ -189,9 +189,23 @@ angular.module('bhima.controllers')
...
 
+    // Restart the search
+    $scope.restartSearch = function () {
+      originalPatientData = null;
+      $scope.patient = { origin_location_id: null, current_location_id: null };
+      $scope.initialOriginLocation = null;
+      $scope.initialCurrentLocation = null;
+
+      // Make sure the patient query gets processed
+      if ('processed' in dependencies.patient) {
```
jniles: There is actually a feature for this.  You can call `validate.refresh(dependencies, ['patient'])` to "refresh" a dependency.  It works for the initial load as well, so if you use the same method for loading and reloading, you can use refresh.

jmcameron: Unfortunately, this does not quite do what is needed.  For the first time a patient is selected, it should process both 'patient' and 'debtorGroup'.  In subsequent calls, it only needs to execute the 'patient' query.    Since there is no need to process the 'debtorGroup' again, it seems like the first time it should do both, but subsequent times only 'patient'.  I don't think adding logic for that is very clean.

I cannot do the valdiate.refresh() call in the restartSearch() function becuase validate.refresh() also calls 'validate.process()' which simply looks up the old data again.   I suppose I could add a function to 'validate' that would clear the selected 'processed' flags.    Suggestions?

PR #310, opened 2015-01-11
Remove has_children field from the unit table

### LINKED ISSUE #267: Issue with the 'unit' table
The 'unit' table has a column 'has_children'.   It seems like this is not necessary since it could be handled by queries?  Steven also noted that even though it seems to be intended as a boolean some rows have odd values for this column.

jniles:
This PR addresses the concerns raised by @jmcameron.  The server side tree controller is restructured to create a unit tree with only a single SQL query, and all client references to has_children has been removed.  Provides SQL migration scripts.

Closes #267


jniles:
@DedrickEnc, could you validate this pull request?   It will allow some future refactoring of db.js.

jniles:
@sfount, considering the revolution in Congo, would you mind merging this?

PR #312, opened 2015-01-12
Account statement num fix

jmcameron:
The account selector on the Report/Account Statement page shows the database ID number in the field once it is selected (before submission).  This mod changes it to display the Account name and number (similar to the drop-down menu during selection).  Once the user presses the [Submit] button, this field is cleared in case the user wants to start another search.

### DIFF
--- a/client/src/partials/reports/account_statement/account_statement.js
+++ b/client/src/partials/reports/account_statement/account_statement.js
@@ -24,6 +24,7 @@ angular.module('bhima.controllers')
     };
     session.config.dateFrom = new Date();
     session.config.dateTo = new Date();
+    session.selectedRequestId = null;
 
     dependencies.accounts = {
       query : {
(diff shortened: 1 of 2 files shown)


client/src/partials/reports/account_statement/account_statement.js
```diff
@@ -24,6 +24,7 @@ angular.module('bhima.controllers')
     };
     session.config.dateFrom = new Date();
     session.config.dateTo = new Date();
+    session.aid = null;
```
jniles: Can this be a more semantic name?

jniles:
Would you mind squashing those merge commits?

jmcameron:
'aid' renamed to 'selectedRequestId'  and commits squashed!

jniles:
Thanks for getting on this so quickly!

PR #314, opened 2015-01-12
Installation/data migration scripts

jniles:
In this PR:
 - migrate.js has been removed.  In its place, a series of shell scripts perform the data dump and alter table queries necessary to upgrade the live database.  A backup dump is made prior to database manipulation.

The intended workflow is as follows:
```bash
$ cd /path/to/bhima/utilities/migrate
$ cat path/to/bhima/server/models/update/*  > sql/upgrade-v1.sql
$ sh 0-run-all-scripts.sh [user] [password] bhima
```
If anything goes amiss during the procedure, the shell scripts are named obviously to aid debugging.


jniles:
@DedrickEnc could you review this PR?

PR #317, opened 2015-01-12
Updates for find patient directive

jmcameron:
Cleaned up translation items for the find patient directive: Removed unused ones, renamed some keys to better reflect the functionality, and removed duplicate items
 
Minor fixes to make it operate more cleanly in both the swap_debtor and patient_edit pages.

### DIFF
--- a/client/src/i18n/en.json
+++ b/client/src/i18n/en.json
@@ -1460,16 +1460,12 @@
     "NO_DISTRIBUTABLE" : "Not distributable"
   },
 
-  "FIND"               : {
-    "ENTER_PATIENT_ID" : "Find by Patient ID",
-    "PATIENT_ID"       : "Patient ID",
-    "PLACEHOLDER"      : "Search patient name",
-    "SEARCH"           : "Find by Name",
-    "TITLE"            : "Find a patient",
-    "ENTER_DEBTOR_ID"  : "Enter Debitor ID",
-    "SEARCH"           : "Search",
-    "PLACEHOLDER"      : "Search debitor",
-    "PATIENT_ID"       : "Patient ID"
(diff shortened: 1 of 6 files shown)


jniles:
While you are working on this, something that has bothered me for a while -- could add the class `btn-sm` to the "submit" button?  It will suddenly fit next to the input, instead of that tiny overflow you can see on it now.

jmcameron:
I was not able to get btn-sm to work.  I tried replacing 'btn' with 'btn-sm' but angular changed it back to 'btn' again on the resulting page.   Not sure why that happened or what to do about it.

I tried hacking the CSS directly in the browser and 'btn-sm' does not really help-- it is too small.  I played around with a new rule that seems to work nice on the form page:
    .btn-sm-bhima {
      padding: 4px 10px;
      margin-left: 2px;
      font-size: 12px;
      line-height: 1.5;
      border-radius: 3px;
    }
 I tried inserting this at the end of form.css, but due to the problem above, I cannot get the class I want into the output html.

jmcameron:
Oopsie.  Figured it out.  I'll add the new rule to form.css and make sure it works.

jniles:
Hmm..  on my machine, the class list "btn btn-sm btn-default" looks perfect.  I'll take a look at this tomorrow with you.

jmcameron:
I'm sure you're right.   It is probably my problem due to the oddities of my quad-res screen setup and adjustments.    So I changed the find patient and account statement buttons to use 'btn btn-sm' and deleted the unneeded btn-sm-bhima CSS rule.  And I squashed commits.

jniles:
This looks great to me now.  Thanks!

PR #319, opened 2015-01-13
Fix small income report bug

jmcameron:
The Incomes report was reporting a javascript error on the startup if no cashbox is selected immediately.   This fix avoids errors due to trying to access undefined values.

### DIFF
--- a/client/src/partials/reports/primary_cash/income/income_report.js
+++ b/client/src/partials/reports/primary_cash/income/income_report.js
@@ -79,23 +79,26 @@ angular.module('bhima.controllers')
     }
 
     function fill () {
-      var request;
+      var request = {
+        dateFrom : util.sqlDate(session.dateFrom),
+        dateTo : util.sqlDate(session.dateTo),
+      };
+
+      // Make sure the account_id has a valid value
       if (session.selectedCash) {
-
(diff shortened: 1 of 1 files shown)


client/src/partials/reports/primary_cash/income/income_report.js
```diff
@@ -78,22 +79,26 @@ angular.module('bhima.controllers')
...
-
-      request = {
+      var request = {
         dateFrom : util.sqlDate(session.dateFrom),
         dateTo : util.sqlDate(session.dateTo),
-        account_id : session.selectedCash.account_id
       };
 
+      // Make sure the account_id has a valid value
+      if (session.selectedCash) {
```
jniles: This will fail if session.selectCash = 0.  Is this the expected behavior?

jmcameron:
I think this is okay.  I just double-checked the code and session.selectedCash is either undefined (when it starts up) or a data object (from the 'data' from the database query).    It never actually contains a number.  I suppose I could add a isNan() check, but that does not seem necessary.

jniles:
Okay, that makes sense.  I didn't scan over the full code to see what values selectCash could take.  Looks good to me.

PR #321, opened 2015-01-13
Sale subsidy integration

DedrickEnc:
In this PR :
- I wrotte a crud module for subsidy
- update patient group
- update database
- update receipt ...


client/src/partials/subsidy/subsidy.js
```diff
@@ -0,0 +1,103 @@
...
+  'connect',
+  'appstate',
+  'uuid',
+  function ($scope, $translate, $http, validate, messenger, connect, appstate, uuid) {
+    var dependencies = {},
+        session = $scope.session = {};
+
+    dependencies.subsidies = {
+      identifier:'uuid',
+      query : '/getSubsidies/'
```
jniles: This query should be done with the validate syntax.

Additionally, it is bad practice to create case-sensitive URLs, as many webservers do not take into account the string case.  This should be avoided in the future.

jniles:
You have quite a few `console.log()`s that should be removed before this goes into development.

client/src/partials/subsidy/subsidy.html
```diff
@@ -0,0 +1,151 @@
+<header data-header>
+  {{ "SUBSIDY.CREATE_TAXE" | translate }}
+</header>
+
+<nav>
+  <div class="pull-left">
```
jniles: Since this module is only a single page, it doesn't make much sense to have breadcrumbs.  What function does it serve?

client/src/partials/subsidy/subsidy.html
```diff
@@ -0,0 +1,151 @@
...
+                <span ng-show="session.new.is_percent">%</span>
+                <span ng-hide="session.new.is_percent">{{enterprise.data[0].symbol}}</span>
+              </span>
+            </div>
+          </div>
+
+          <div class="form-group">
+            <label class="control-label col-xs-4 required">{{ "COLUMNS.DEBITOR_GROUP" | translate }}</label>
+            <div class="col-xs-7">
+              <select class="form-bhima" ng-model="session.new.debitor_group_uuid">
```
jniles: This list should be sorted.

client/src/partials/subsidy/subsidy.html
```diff
@@ -0,0 +1,151 @@
...
+        <div class="panel-heading">
+          <span class="glyphicon glyphicon-briefcase"></span>
+          {{ "SUBSIDY.REGISTERED" | translate }}
+        </div>
+        <table class="table table-condensed">
+          <thead>
+            <tr>
+              <th>{{ "COLUMNS.LABEL" | translate}}</th>
+              <th>{{ "COLUMNS.VALUE" | translate}}</th>
+              <th>{{ "COLUMNS.IS_PERCENT" | translate}}</th>
```
jniles: Rather than "is percent" as a column, we should show values like "10%" or "10$".  It is much easier to see, and will free up some more space in this module.

jniles:
This PR is broken.  Steps to reproduce:
1) Create a patient with out a patient group.
2) Try to bill him
3) The server tries to execute `INSERT INTO sale_subsidy () VALUES ;`, printed in the console.

DedrickEnc:
Thank you for your feedback, I have pushed some commits about the two major issues and I will create a specific branch to let every one contribute.

client/src/partials/subsidy/subsidy.js
```diff
@@ -0,0 +1,103 @@
...
+        connect.delete('subsidy', 'uuid', [subsidy.uuid])
+        .then(function () {
+          $scope.subsidies.remove(subsidy.uuid);
+          messenger.info($translate.instant('SUBSIDY.DELETE_SUCCESS'));
+        });
+      }
+    };
+
+    $scope.edit = function (subsidy) {
+      console.log('le voici', subsidy);
```
jniles: Remove unnecessary `console.log()`s.

client/src/partials/subsidy/subsidy.js
```diff
@@ -0,0 +1,103 @@
...
+    $scope.new = function () {
+      session.action = 'new';
+      session.new = {};
+    };
+
+    $scope.save = {};
+
+    $scope.save.edit = function () {
+
+      var record = angular.copy(connect.clean(session.edit));
```
jniles: connect.clean() will do essentially the same thing as angular.copy().  You can remove the angular.copy() call.

server/controllers/createSale.js
```diff
@@ -89,3 +97,33 @@ function generateSaleItems(saleRecordId, saleItems) {
...
+      }
+      currentCost = currentCost - amount;
+      return {uuid : uuid(), sale_uuid : saleRecord.uuid, subsidy_uuid : item.uuid, value : amount};
+    }
+  });
+
+  var saleSubsidies = subsidyList.filter(function (item){
+    return item.value > 0;
+  });
+  return parser.insert('sale_subsidy', saleSubsidies);
```
jniles: In the past we have used parser on the server side, but doing so makes it really hard to read.  It also will be difficult to migrate to other frameworks going forward.  Finally, it takes twice as much processing on the server, since the server must 1) make a JSON object and 2) interpret the JSON object into a string with escapes and 3) execute it.

If you compose your queries as a string, the server must only concatenate and execute the string.  Is there a way you can easily transform this into a string?

server/controllers/journal.js
```diff
@@ -238,6 +235,17 @@ function handleSales (id, user_id, done, caution) {
...
 
+
+  function getSubsisy () {
+    var sql =
+    'SELECT `sale_subsidy`.`value`, `debitor_group`.`account_id`, `subsidy`.`text`, `sale`.`uuid` FROM ' +
+    '`sale_subsidy`, `debitor_group`, `subsidy`, `sale` WHERE `sale_subsidy`.`sale_uuid`=`sale`.`uuid` AND ' +
+    '`sale_subsidy`.`subsidy_uuid`=`subsidy`.`uuid` AND `subsidy`.`debitor_group_uuid`=`debitor_group`.`uuid` AND ' +
+    '`sale_subsidy`.`sale_uuid`=' + sanitize.escape(id);
+
+    return db.exec(sql);
```
jniles: You should use db.exec()'s built-in escapes.  Instead of `sale_uuid=' +sanitize.escape(id)`, you can write `sale_uuid=?;'` and then pass the id to `db.exec(sql, [id])`.

jniles:
Since this is on a separate branch, I'm closing this PR.

PR #327, opened 2015-01-13
Update globaltransaction

lomamech:
Display transactions with their original currencies
Amelioration of presentation of this reports


client/src/partials/reports/global_transaction/global_transaction.html
```diff
@@ -70,10 +64,9 @@
                 <td width="20%"> {{ record.trans_id }} </td>
```
jniles: @lomamech , 20% + 20% + 20% + 20% + 20% + 50% = 150%.  Please correct your widths so that the text is visible!

client/src/partials/reports/global_transaction/global_transaction.html
```diff
@@ -107,10 +100,9 @@
...
```
jniles: 20% + 20% + 20% + 20% + 20% + 60% = 160%. Please correct your widths so that the text is visible!

client/src/partials/reports/global_transaction/global_transaction.js
```diff
@@ -133,45 +142,64 @@ angular.module('bhima.controllers')
...
+        datet : util.sqlDate($scope.dates.to)
       };
 
-      if ($scope.model.account_id && $scope.model.account_id === 0) {
-        $scope.model.account_number = 'Tous';
-      } else {
-        $scope.model.account_number = $scope.accounts.get($scope.model.account_id).account_number;
-      }
-
       connect.fetch('/reports/allTrans/?'+JSON.stringify(qo))
```
jniles: Could you make this route more semantic?  Something like `/reports/transactions`>

client/src/partials/reports/global_transaction/global_transaction.js
```diff
@@ -133,45 +142,64 @@ angular.module('bhima.controllers')
...
-      $scope.mode = $scope.model.account_id !== 0 ? 'selected' : 'all';
+      $scope.mode = ($scope.model.account_id && $scope.model.account_id > 0) ? 'selected' : 'all';
+      
+      if ($scope.model.account_id && $scope.model.account_id > 0) {
+        $scope.model.account_number = $scope.accounts.get($scope.model.account_id).account_number;       
+      } else {
+        $scope.model.account_number = 0;
+      }
+
       var qo = {
```
jniles: Putting things in a JSON object like this is unnecessary, this will generate a URL in this format: `/reports/allTrans/?{key:"value",key2:"value2"}`.  In the server logs, this is difficult to read.  It will we better if you composed these as parameters: `/reports/allTrans?key=value&key2=value2`.

client/src/partials/reports/global_transaction/global_transaction.js
```diff
@@ -133,45 +142,64 @@ angular.module('bhima.controllers')
     }
 
     function search () {
-      if (!$scope.model.account_id) { return; }
-      $scope.mode = $scope.model.account_id !== 0 ? 'selected' : 'all';
+      $scope.mode = ($scope.model.account_id && $scope.model.account_id > 0) ? 'selected' : 'all';
```
jniles: Since you use `$scope.model.account_id && $scope.model.account_id > 0` in more than one place, it would be good to assign it to a local variable.  Something like this:
```javascript
var hasSelectedAccount = $scope.model.account_id && $scope.model.account_id > 0;
$scope.mode = hasSelectedAccount ? 'selected' : 'all';
if (hasSelectedAccount) {
  // etc..
}
```

client/src/partials/reports/global_transaction/global_transaction.js
```diff
@@ -133,45 +142,64 @@ angular.module('bhima.controllers')
...
 
-      if ($scope.model.account_id && $scope.model.account_id === 0) {
-        $scope.model.account_number = 'Tous';
-      } else {
-        $scope.model.account_number = $scope.accounts.get($scope.model.account_id).account_number;
-      }
-
       connect.fetch('/reports/allTrans/?'+JSON.stringify(qo))
       .then(function (res) {
         if (res.length > 0) {
```
jniles: You do not need to check if res.length > 0.  If res is an empty array, `res.map()` will not be called.

client/src/partials/reports/global_transaction/global_transaction.js
```diff
@@ -133,45 +142,64 @@ angular.module('bhima.controllers')
...
         }
       });
     }
 
     function getTotal(items) {
+      var sCredit = 0, 
+        sDebit = 0;
       $scope.somCredit = 0;
       $scope.somDebit = 0;
       if (items.length > 0) {
```
jniles: Just like above, you do not need to check `items.length > 0` before calling items.forEach().  If items is an empty array, .forEach() will not run.

If you are running into errors where items in undefined, you should check for that instead.  Use `angular.isDefined(items)` as your check.

server/controllers/report.js
```diff
@@ -607,16 +607,34 @@ function expenseReport (params) {
...
+      '`uuid`, ', '`inv_po_id`, ',
+      '`trans_date`, ', '`debit_equiv`, ',
+      '`credit_equiv`, ', '`account_id`, ',
+      '`deb_cred_uuid`, ', '`currency_id`, ', '`doc_num`, ',
+      '`trans_id`, ', '`description`, ', '`comment` '
+    ].join(source[params.source] + '.');
+    sub_chaine = source[params.source] + '.' + sub_chaine;
+
+    requette =
+      'SELECT `t`.`uuid`, `t`.`trans_id`, `t`.`trans_date`, `ac`.`account_number`, `t`.`debit_equiv` AS `debit`,  ' +
```
jniles: Use a semantic name, instead of "t".  You can also remove the "`" tick marks, they do not do anything.

client/src/partials/reports/global_transaction/global_transaction.js
lomamech: I have use reports/allTransactions because in file reports on server there are yet transactions

client/src/partials/reports/global_transaction/global_transaction.js
```diff
@@ -77,22 +80,19 @@ angular.module('bhima.controllers')
         datef : util.sqlDate($scope.dates.from),
         datet : util.sqlDate($scope.dates.to)
       };
+      var url = '/reports/allTransactions/?source=' + $scope.model.source_id +'&enterprise_id=' + $scope.enterprise.id + '&account_id=' + $scope.model.account_id + '&datef=' + util.sqlDate($scope.dates.from) + '&datet=' + util.sqlDate($scope.dates.to);
```
jniles: This is much better!  It will be more readable if you use multiple lines:
```javascript
var url = '/reports/allTransactions/?source=' + $scope.model.source_id +
   '&enterprise_id=' + $scope.enterprise.id + 
   '&account_id=' + $scope.model.account_id + 
   '&datef=' + util.sqlDate($scope.dates.from) + 
   '&datet=' + util.sqlDate($scope.dates.to);
```

client/src/partials/reports/global_transaction/global_transaction.js
```diff
@@ -135,10 +131,9 @@ angular.module('bhima.controllers')
     }
 
     function search () {
-      console.log('Le modele de la source ID',$scope.model.source_id);
       $scope.mode = ($scope.model.account_id && $scope.model.account_id > 0) ? 'selected' : 'all';
```
jniles: You could also use hasSelectedAccount here to clean the code up more.

client/src/partials/reports/global_transaction/global_transaction.js
```diff
@@ -151,33 +146,43 @@ angular.module('bhima.controllers')
...
+        if($scope.enterprise.currency_id !== item.currency_id){
+          sCredit += item.credit / exchange.rate(item.credit,item.currency_id,new Date()); 
+          sDebit += item.debit / exchange.rate(item.debit,item.currency_id,new Date());
+        } else {
+          sCredit += item.credit;
+          sDebit += item.debit;
+        }
+                
+        if($scope.enterprise.currency_id === $scope.model.c){
+          $scope.somCredit = sCredit;
```
jniles: $scope.somCredit is updated with every iteration of the .forEach() loop.  Are you sure this is what you want to do?

server/controllers/report.js
```diff
@@ -605,52 +605,52 @@ function expenseReport (params) {
   return defer.promise;
 }
 
-function allTrans (params){
+function allTransactions (params){
```
jniles: Much better!

client/src/partials/reports/global_transaction/global_transaction.js
```diff
@@ -57,7 +58,9 @@ angular.module('bhima.controllers')
         account.account_number = String(account.account_number);
       });
       $scope.model.c = $scope.enterprise.currency_id;
```
jniles: Why not change `$scope.model.c` to `$scope.currencyId`?  They are equal length (to type), and $scope.currencyId is one less level deep in the $scope object.  Also, $scope.currencyId is much more readable.

server/controllers/report.js
```diff
@@ -605,52 +605,52 @@ function expenseReport (params) {
...
   var source = {
     '0' : 0,
     '1' : 'posting_journal',
     '2' : 'general_ledger'
   };
   var def = q.defer();
-  params = JSON.parse(params);
+  params = querystring.parse(params);
```
jniles: Yes!  Much better.

jniles:
This PR looks good to me.  Are you done with all your fixes, or should I wait to merge it in?

lomamech:
You can merge it

2015-01-14 10:31 GMT+01:00 Jonathan Niles <notifications@github.com>:

PR #328, opened 2015-01-14
Feature/ui stock entry

mbayopanda:


client/src/partials/stock/entry/report.js
```diff
@@ -47,6 +47,7 @@ angular.module('bhima.controllers')
     function startup(models) {
       angular.extend($scope, models);
 
+      session.depot_entry = $scope.receipt.data.length ? $scope.receipt.data[0].depot_entry : null;
```
jniles: This should be mixedCase.

PR #331, opened 2015-01-14
Fix: Currencies changing error

lomamech:
Fix: Currencies changing error

### DIFF
--- a/client/src/partials/receipts/receipt.sale.js
+++ b/client/src/partials/receipts/receipt.sale.js
@@ -70,20 +70,25 @@ angular.module('bhima.controllers')
 
     function updateCurrency (currency) {
       var totals = model.total,
-        saleRecords = model.saleRecords[0];
-        
-      totals.localeCost = model.common.doConvert(saleRecords.cost, currency, saleRecords.invoice_date);
+        saleRecord = model.saleRecords,
+        saleRecords = model.saleRecords[0],
+        doConvert = model.common.doConvert,
+        convert = model.common.convert,
+        ledgers = model.ledger;
 
-      if (model.ledger)  {
-        totals.localeBalance = model.common.doConvert(model.ledger.balance, currency, saleRecords.invoice_date);
-        model.ledger.localeCredit = model.common.doConvert(model.ledger.credit, currency, saleRecords.invoice_date);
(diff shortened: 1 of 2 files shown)


client/src/partials/receipts/receipt.sale.js
```diff
@@ -87,7 +104,6 @@ angular.module('bhima.controllers')
     }
 
     function updateCost(currency_id) {
```
jniles: Is this function still used?

client/src/partials/receipts/receipt.sale.js
```diff
@@ -67,6 +68,22 @@ angular.module('bhima.controllers')
       }
     };
 
+    function updateCurrency (currency) {
+      model.total.localeCost = model.common.doConvert(model.saleRecords[0].cost, currency, model.saleRecords[0].invoice_date);
```
jniles: You should really cache this in a local variable.  Like this:
```javascript
var totals = model.total.localCost;
// do something fancy
```
This is both semantic and faster.  The performance increases [are](http://stackoverflow.com/questions/2936807/javascript-object-access-performance) [well](http://moduscreate.com/javascript-performance-tips-tricks/) [documented](http://www.javascriptkit.com/javatutors/efficientjs.shtml).

client/src/partials/receipts/receipt.sale.js
lomamech: Yes, This function is used when downloading page

client/src/partials/receipts/receipt.sale.js
```diff
@@ -67,6 +68,25 @@ angular.module('bhima.controllers')
       }
     };
 
+    function updateCurrency (currency) {
+      var totals = model.total,
+        saleRecords = model.saleRecords[0];
+        
+      totals.localeCost = model.common.doConvert(saleRecords.cost, currency, saleRecords.invoice_date);
```
jniles: The same logic as totals = model.total can be applied to almost every line of this function.  For example, you call `model.common.doConvert()` potentially 4 + length(model.saleRecords) times in updateCurrency()!  You clearly should cache this function in a local variable that can be called less expensively.

PR #333, opened 2015-01-14
Created new find account directive

jmcameron:
Created a new findAccount directive.

Updated the Report/Account Statement page to use the new directive.

I've added several options to the findAccount directive that should make it more useful in various contexts.  I addeda  documentation page in the client/src/js/directives folder:

  [findaccount.md](https://github.com/jmcameron/bhima/blob/feature/find_account_directive/client/src/js/directives/findaccount.md)

### DIFF
--- a/client/src/i18n/en.json
+++ b/client/src/i18n/en.json
@@ -883,6 +883,11 @@
    "WEEK"         : "This Week"
    },
 "FIND": {
+    "ACCOUNT" : {
+      "ACCOUNT"         : "Account",
+      "ENTER_ACCOUNT"   : "Enter Account",
+      "PLACEHOLDER"     : "Account Name or Number"
+      },
    "PATIENT_ID"        : "Patient ID",
    "PATIENT_NAME"      : "Patient name",
    "SEARCH_NAME"       : "Search by Name",
--- a/client/src/i18n/fr.json
+++ b/client/src/i18n/fr.json
@@ -883,6 +883,11 @@
    "WEEK"         : "Cette semaine"
    },
 "FIND": {
+    "ACCOUNT" : {
+      "ACCOUNT"         : "Compte???",
+      "ENTER_ACCOUNT"   : "Entrez Compte???",
+      "PLACEHOLDER"     : "Compte Nom ou Numéro???"
+      },
    "PATIENT_ID"        : "ID Patient",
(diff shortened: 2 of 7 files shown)


jniles:
The code looks good to me. If someone else can verify that it behaves as expected, we can pull this in.

jmcameron:
@DedrickEnc Please review this PR and if it is okay, merge into development.

jniles:
@jmcameron It seems like the directive is a bit buggy.  Once a selection is chosen, you can search for another selection just fine ... however all the titles displayed are the selected account.  I've attached a screenshot to illustrate.

accountdirective

jmcameron:
Strange... I am not seeing this issue at all.  You are using my branch feature/find_account_directive and it is up to date, right?

I tried searching for new accounts several times in a row and it worked fine each time for me.

jniles:
Alright, I'll debug it on my machine and let you know.

jniles:
More details:
I am using the latest pull from your branch.  On the report page, you can still use the text input after clicking "submit".  Is this expected behavior?  If you try to do execute a new search without hitting the "refresh" button, the screen shot is the result.

However, upon refresh, the whole directive works correctly.  My recommendation is that we either disable the search after an initial search, or we remove the "refresh" functionality and make the directive behave like the other "find" directives.

Can you reproduce the bug on your end?

jmcameron:
Ok, I understand.  I never thought of that.  I always pressed the refresh button first.  I think I can fix that pretty easity.

jmcameron:
I've fixed this issue.  Now after you select account (before you press [Submit]), the field becomes read-only.  So to select another account, you must press the refresh button.   And once you have pressed the [Submit] button, the field is cleared and the next search should work fine.    I played around with getting the form to clear itself if you click in the account field after an account is selected, but I could not get that to work properly.  I think this approach is reasonable.

NOTE:  Both my feature/edit_budget PR and this PR have exactly the same code relating to the new findAccount directive and the adaptation of the account statement report to use it.  Therefore, if the feature/edit_budget branch is merged in first, this PR can be closed.    If this PR is merged first, the feature/edit_budget branch is still needed even though it does include the changes in this PR.  The changes are identical, so I do not think it will cause any merge conflicts.

jniles:
Alright, this behaves as expected for me.

PR #335, opened 2015-01-15
Fixing ui editing subsidy

mbayopanda:


DedrickEnc:
Thank you for the pull request, do your best to provide a quick description next time.

client/src/partials/subsidy/subsidy.html
```diff
@@ -125,8 +129,12 @@
 
           <div class="form-group">
             <label class="control-label col-xs-4 required">{{ "COLUMNS.VALUE" | translate }}</label>
-            <div class="col-xs-7">
+            <div class="input-group col-xs-7">
               <input ng-model="session.edit.value" class="form-bhima" required="required">
+              <span class="input-group-addon">
+                <span ng-show="session.edit.is_percent">%</span>
+                <span ng-hide="session.edit.is_percent">{{enterprise.data[0].symbol}}</span>
```
jniles: Shouldn't this be tied to the enterprise the user is currently using?  It seems like if we have more than one enterprise with different currency symbols, this will be incorrect.

PR #336, opened 2015-01-15
Added ability to handle IN expressions in WHERE clauses.

jmcameron:
Added ability to handle IN expressions in WHERE clauses.   
Added a couple of tests to validate correct operation.


jniles:
Looks good to me.

PR #337, opened 2015-01-15
Fixing price list css

mbayopanda:
In this PR : 
- Fxing css in 'Admin/Price list' when selecting an item


jniles:
@mbayopanda, rather than introduce a new class "white", why not just make sure all glyphicons that are children of .selected-primary are white?  A rule like `.selected-primary .glyphicon { color : #fff; }` seems like a much better choice to me.

PR #346, opened 2015-01-17
Parser escapes MySQL keywords

### LINKED ISSUE #334: Error when insert or update currency
I got error message :
[CONNECT] Internal Server Error
Status Code 500 - "Internal Server Error"
Description
Server returned a '500' internal error response for url '/data/'.
Parameters

thanks

jniles:
This PR closes #334.  Both the UPDATE and INSERT functions now escape mysql keywords.  Additional small fixes to the currency module.  Test written for each case.

It is my recommendation that we remove the ability to add currencies on the fly.  Currencies might be set up in an additional installation step.  Having the system as flexible as it is currently seems _very_ unsafe.


jniles:
__NOTE__: The code in this PR is rather crude, since it is a quick fix.  If we do intend on keeping these fixes long term, I'll do a bit of refactoring and pull in some of the UI adjustments @lomamech did in his branch.

PR #351, opened 2015-01-27
Harmonisation of interfaces of the stock modules and some reports

lomamech:
Harmonization of interfaces of the stock modules and some reports


jniles:
I noticed that many of these controllers have nearly identical code...  Is it possible to make a single generic controller to process some, if not all, of these receipts?

PR #352, opened 2015-01-27
Lock employee

lomamech:
Allow the possibility to lock the status of an employee, in all system modules


client/src/partials/journal_voucher/journal_voucher.js
```diff
@@ -63,9 +63,11 @@ angular.module('bhima.controllers')
         'tables' : {
```
DedrickEnc: This query does not allow user to add a transaction for the supplier, only for the employee.
Please correct it.

jniles:
What is a "locked" employee?  It is the same as firing an employee?  Or is it some other condition?

DedrickEnc:
In some situation an employee must not be concerned in some transaction, or he can die or be fired, or just not available. In this case, we must lock him.

This functionality is important for multi payroll, because you can select employees who will not be payed by locking them.

PR #370, opened 2015-01-29
implementing balance sheet continues

DedrickEnc:
In this PR, we created a set of user interface to generate the balance sheet report.


jmcameron:
Does this duplicate the income/expense account budget/balance page in the feature/budget page?

DedrickEnc:
No, this is just a pdf report. these are two different things.

PR #401, opened 2015-02-12
Restoration of buttom submit for the directive found account

lomamech:
Restoration of buttom submit for the directive found account


lomamech:
Thank Jonathan Cameron, i will found another way for select an account for account Statement

Chris

jmcameron:
Very good.

I will work on an extra option (attribute) that will tell the findAccount directive not to display a Submit button so that we can use it in both places nicely.

Also, I noticed that the budget-related translation items in the French (fr.json) have not been translated into French.   One of you should search that file for '???' and replace them with French translations.  

I know you guys have been busy, but I encourage you to try the Budget pages to make sure you understand them and that they work for you.   If there are problems with these two pages, I can help fix them.

-Jonathan

PR #403, opened 2015-02-14
Update the findAccount directive to support hiding its submit button

jmcameron:
This updates the findAccount directive so it supports both the Budget>Edit Budget page (with its own [Submit] button) and the Reports>Account Statement page (with a separate [Submit] button).

Basically, the findAccount now supports an additional directive to hide the [Submit] button on the right end of the findAccount field:

      hide-submit

I've used this on the Report>Account Statement page to hide the findAccount's submit button there.

NOTE: There is documentation for this directive in the Markdown text file 'client/src/js/directives/findaccount.md' which you can see nicely formatted here:

   https://github.com/IMA-WorldHealth/bhima/blob/development/client/src/js/directives/findaccount.md

I have not addressed the 'Select Account', 'Find Account' duplication on the Account Statement page.  One of you can address that.


jmcameron:
I just added the "folder open" glyphicon on the left end of the findAccount directive to make it more consistent with the way that it used to be in the Accounts statement page.

PR #406, opened 2015-02-16
Adding a new directive for found account without a submit button

lomamech:
Adding a new directive for found account without a submit button


jmcameron:
Is this needed now that I added the 'hide-submit' button in the findAccount directive?

PR #500, opened 2015-04-28
Many changes about database

DedrickEnc:


utilities/account/integration/ohada.csv
```diff
@@ -0,0 +1,1193 @@
...
+,7588,Autres Produits divers
+,75881010,Autres revenus
+,759,REPRISES DE CHARGES PROVISIONN�ES D'EXPLOITATION
+,7593,Reprises de charges provisionnees sur stocks
+,75931010,Reprise des charges provisionnees pour stocks obsol�te
+,7594,Reprises de charges provisionnees sur creances
+,75941010,Reprise de charges provisionnees sur creances impayees
+,7598,Reprises sur autres charges provisionnees
+,77,REVENUS FINANCIERS ET PRODUITS ASSIMIL�S
+,771,INT�R�TS DE PR�TS
```
jniles: Is there any way to remove the code points that are causing these ? to appear?  They will probably not be inserted correctly into the database.

DedrickEnc: I am Trying to avoid that at the next pull request

PR #511, opened 2015-04-30
Rework logger for server

jniles:
This PR removes the logger.js and associated plugins from the server architecture.  For production environments, we should use a production-quality logger such as [morgan](https://github.com/expressjs/morgan).  Support for morgan will be added in a future commit.


jniles:
@DedrickEnc , can you approve these changes?

DedrickEnc:
I will see that

jniles:
Will this even make it into production?  Asking for a friend.

DedrickEnc:
It will be merged in the development branch after the trip.

PR #514, opened 2015-05-02
Code cleanup/reports

jniles:
In this PR:
  - Removed rt.js, which was never used
  - misc cleanup in reports.js

There's a lot of work to be done in the reports module, but it requires a working database to make sure I do not break anything.  Once the OHADA changes are done, I'll resume cleaning.  Until then, this PR can safely be integrated into the main repository.


DedrickEnc:
Thank you Jon, I think you are very sure that rt.js is never used. and I will integrate OHADA into the system today. thanks.

jniles:
@DedrickEnc , rt.js was never used - I wrote it with a terrible API which @sfount pointed out to me.  So, we wrote report.js to do everything rt.js did and more.  All routes were migrated away from it last fall, unless someone tried to use it since then.

PR #519, opened 2015-05-06
Updates creditor_group, debitor_group, patient_registration

mbayopanda:
In this PR, we have : 
- Added the support of only ohada account in creditor_group and debitor_group management
- Adding a machanism to forbid a duplication of a patient according his hospital_number ('hospital_no')

### DIFF
--- a/client/src/partials/patient_registration/patient.js
+++ b/client/src/partials/patient_registration/patient.js
@@ -64,7 +64,8 @@ angular.module('bhima.controllers')
     dependencies.debtorGroup = {
       query : {
         identifier : 'uuid',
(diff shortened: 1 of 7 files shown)


client/src/partials/patient_registration/patient.js
```diff
@@ -79,11 +80,32 @@ angular.module('bhima.controllers')
...
+      var query = {
+        tables : { 
+          patient : { columns : ['uuid'] }
+        },
+        where  : ['patient.hospital_no=' + file_number]
+      };
+      connect.fetch(query)
+      .then(function (res) {
+        console.info(res);
+        if (res.length === 0) {
```
DedrickEnc: I think the code is very clear however it is less nice. lol

PR #520, opened 2015-05-08
Fixing zero file_number

mbayopanda:
In this PR we have fixed the zero file number when register a patient

### DIFF
--- a/client/src/partials/patient_registration/patient.js
+++ b/client/src/partials/patient_registration/patient.js
@@ -86,6 +86,10 @@ angular.module('bhima.controllers')
     }
 
     function checkingExistPatient (file_number) {
+      if (file_number === 0) {
+        return false;
+      }
+
       var def = $q.defer();
       var query = {
         tables : { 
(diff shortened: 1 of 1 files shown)


client/src/partials/patient_registration/patient.js
```diff
@@ -86,6 +86,10 @@ angular.module('bhima.controllers')
     }
 
     function checkingExistPatient (file_number) {
```
jniles: If I understand this code properly, if `file_number === 0`, `checkExistPatient()` returns a boolean, otherwise it returns a promise.  Wouldn't it be better to use `def.resolve(false)`?  That way you are return a consistent type.

client/src/partials/patient_registration/patient.js
```diff
@@ -95,13 +99,8 @@ angular.module('bhima.controllers')
...
       connect.fetch(query)
       .then(function (res) {
-        console.info(res);
-        if (res.length === 0) {
-          def.resolve(false);
-        } else {
-          def.resolve(true);
-        }
-      })
+        def.resolve(res.length === 0);
```
jniles: :)

PR #525, opened 2015-05-12
Fix credit notsubvention

lomamech:
Update sql request for credit not, fix error for subsidy


server/controllers/journal.js
```diff
@@ -1104,57 +1105,72 @@ function handleCreditNote (id, user_id, done) {
...
+      item.description = cfg.descrip;
+      item.period_id = cfg.periodId;
+      item.fiscal_year_id = cfg.fiscalYearId;
+      item.trans_id = cfg.trans_id;
+      item.trans_date = util.toMysqlDate(get.date());
 
-    return db.exec(queries.debtor);
-  })
-  .then(function () {
+      //FIX ME deb_cred_uuid when is empty the text 'null' is inserted in the table posting_journal
```
jniles: If you use `db.exec(sql, [param1, param2, ...])` instead of sanitize.escape, you can db.js will replace a the javascript object `null` will the a NULL database value.  For more information, see this [link](https://github.com/felixge/node-mysql/#escaping-query-values)

PR #529, opened 2015-05-12
Update report of sale record and auxillary cash report

lomamech:
Update report of sale record and auxillary cash report

### DIFF
--- a/client/src/partials/records/auxillary_cash_records/auxillary_cash_recors.js
+++ b/client/src/partials/records/auxillary_cash_records/auxillary_cash_recors.js
@@ -165,10 +165,16 @@ angular.module('bhima.controllers')
     }
 
     function convert (){
+      console.log('PPPPPPPPPPPPPPPPPPPPPP');
+      console.log($scope.model.cash.data);
+      console.log('PPPPPPPPPPPPPPPPPPPPPP');
+
       session.sum = 0;
       if($scope.model.cash.data) {
         $scope.model.cash.data.forEach(function (cash) {
-          session.sum += exchange.convertir(cash.cost, cash.currency_id, session.currency, new Date());
+          if(!cash.cashDiscardId){
+            session.sum += exchange.convertir(cash.cost, cash.currency_id, session.currency, new Date());
+          }
         });
       }
     }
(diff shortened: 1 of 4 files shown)


client/src/partials/records/auxillary_cash_records/auxillary_cash_recors.js
```diff
@@ -165,10 +165,16 @@ angular.module('bhima.controllers')
     }
 
     function convert (){
+      console.log('PPPPPPPPPPPPPPPPPPPPPP');
```
jniles: Why are you committing these `console.log`'s the development branch?  Are they necessary?

client/src/partials/records/sales_records/sales_records.js
```diff
@@ -142,7 +142,12 @@ angular.module('bhima.controllers')
 
     function totalCost() {
       return $scope.model.sale.data.reduce(function (a, b) {
```
jniles: It might be better to filter the model first, so that you are not adding `0` without reason.  Something like:
```
var items = $scope.model.sale.data;
var total = items.filter(function (item) { return !item.creditId; })
.reduce(function (a, b) { return a + b.cost; }, 0);
return total;
```

DedrickEnc: var items = $scope.model.sale.data;
var total = items.filter(function (item) { return !item.creditId; })
.reduce(function (a, b) { return a + b.cost; }, 0);
return total;

This instruction is good, but not fast because the array will be covered twice, for the filter function and for the reduce function.

client/src/partials/records/auxillary_cash_records/auxillary_cash_recors.js
lomamech: I just delete it in a new branch, I forgot to do it

client/src/partials/records/sales_records/sales_records.js
jniles: True, but I am more worried about code clarity at this point than speed.  Both the current loop and my suggested loop are `O(n)`, which are theoretically the same speed.

client/src/partials/records/auxillary_cash_records/auxillary_cash_recors.js
jniles: The purpose of pull requests is to fix these issues *before* they land in the main repository.  It would have been nice to fix this simple error before merging into development.

DedrickEnc:
You are right, the complexity is polynomial in this case, but I think in reality the first proposition takes more time. However you are right to be worried for the code clarity but we have also to be worried about the reality speed.
lol

PR #541, opened 2015-05-19
Improve management of account

lomamech:
Improve management of account


client/src/partials/accounts/create_account/create.js
```diff
@@ -162,7 +163,9 @@ angular.module('bhima.controllers')
         is_ohada    : $scope.editAccount.is_ohada,
         locked      : $scope.editAccount.locked,
         cc_id       : $scope.editAccount.cc_id,
-        pc_id       : $scope.editAccount.pc_id
+        pc_id       : $scope.editAccount.pc_id,
```
jniles: Why do you have this property set twice?

jniles:
LGTM

PR #543, opened 2015-05-19
Allow report of sale regristration and cash paiement by User

lomamech:
Allow report of sale regristration and cash paiement by User

### DIFF
--- a/client/src/i18n/en.json
+++ b/client/src/i18n/en.json
@@ -132,6 +132,8 @@
   },
 "AUXILLARY_CASH_RECORD" : {
   "ALL_PROJECTS"    : "All projects",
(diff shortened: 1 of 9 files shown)


client/src/i18n/en.json
```diff
@@ -132,6 +132,8 @@
   },
 "AUXILLARY_CASH_RECORD" : {
   "ALL_PROJECTS"    : "All projects",
+   "ALL_USERS"      : "All Users",
```
jniles: I've noticed that in several pull requests, you guys add the same property several times in en.json and fr.json.  It might be better to move these shared properties to a different namespace, such as "RECORDS.ALL_USERS", that way we don't fill up our JSON files with repeated property definitions.

PR #549, opened 2015-06-10
Unit reference

lomamech:
Adding unit and module management for reference and reference group


client/src/partials/reference_group/reference_group.js
```diff
@@ -0,0 +1,126 @@
...
+    appstate.register('enterprise', function (enterprise) {
+      $scope.enterprise = enterprise;
+      validate.process(dependencies)
+      .then(startup);
+    });
+
+    $scope.delete = function (reference_group) {
+      var result = confirm($translate.instant('REFERENCE_GROUP.CONFIRM'));
+      if (result) {  
+        connect.basicDelete('reference_group', reference_group.id, 'id')
```
jniles: Do not use `connect.basicDelete()`, it is deprecated.  You should use `connect.delete()` instead.

client/src/partials/reference_group/reference_group.js
```diff
@@ -0,0 +1,126 @@
...
+        session.action = '';
+        session.edit = {};
+      });
+    };
+
+    $scope.save.new = function () {
+      var record = connect.clean(session.new);
+      record.is_report = (session.new.is_report)?1:0;
+
+      connect.basicPut('reference_group', [record])
```
jniles: Use `connect.post()`.

client/src/partials/reference_group/reference_group.js
```diff
@@ -0,0 +1,126 @@
...
+        messenger.success($translate.instant('REFERENCE_GROUP.SAVE_SUCCES'));        
+        record.reference = generateReference(); 
+        $scope.reference_groups.post(record);
+        session.action = '';
+        session.new = {};
+      });
+    };
+
+    function generateReference () {
+      window.data = $scope.reference_groups.data;
```
jniles: Why are you attaching things to the `window` object?  If you need to use the browser window, you should import angular's $window service.
```js
angular.controller('someController', ['$window', function ($window) {
  // use '$window' where you would use 'window'
  // ...
});
```
It is safer.  See the documentation [here](https://docs.angularjs.org/api/ng/service/$window).

client/src/partials/reference/reference.js
```diff
@@ -0,0 +1,138 @@
+angular.module('bhima.controllers')
+.controller('reference', [
+  '$scope',
+  '$translate',
+  '$http',
+  'validate',
+  'messenger',
+  'connect',
+  'appstate',
+  function ($scope, $translate, $http, validate, messenger, connect, appstate) {
```
jniles: Why inject `$http`?  You  never use it in this controller.

PR #554, opened 2015-06-16
New plugin manager; complete mail plugin rewrite

jniles:
This PR implements a new plugin manager (plugins/pluginManager.js).  It is fairly bare-bones at this point, but the structure is fully functional.  This PR is mostly for review by the team, with additions to be made in the future.

Each plugin is loaded in a separate thread using `child_process.fork([plugin])`.  Configuration parameters are defined in `config/environment/server.js`.  Each plugin must define a schedule using a crontab syntax.   The schedule is parsed and interpreted using [laterjs](http://bunkat.github.io/later/).

The mail plugin has been completely rewritten as an example plugin.  Currently, the daily report has been implemented, but it should be very easy to create new reports by putting a folder in `mail/reports/{report_name}/` with `queries.json`, `{report_name}.tmpl.html`, and `{report_name}/lang/en.json` or `{report_name}/lang/fr.json` with appropriate translations.  Note: the French daily report is not yet translated.  For more information on how this works, see the README (`plugins/mail/README.md`).

A nice bonus with the mail plugin is that templates are now defined using handlebarsjs-like syntax.  You can now write:
```html
<h1> {{ this.data.patientCount }} </h1>
```
and use a JSON object `{ data : { patientCount : 12 } }` to render the template with the correct values.  For database queries, specifying the locality ("en" or "fr") of the currency in the query will lead to automatic currency formatting in either Congolese Francs or USD.  The code for this is located inside `plugins/mail/lib/locales.js`.

Any feedback would be appreciated.  There are still some bugs to be patched, but this is my proposal for the future of bhima plugins.  In the future, I'd like to move plugins into separate repositories, making them easier to maintain.  Any comments would be welcome.


DedrickEnc:
It seems interest, I will read that and give some suggestion eventually. I will merge it soon.
For now we have symmetricDS, mail, backup, and SNIS as plugins, if every things are okay with this PR, we should find a way to follow this way.

jniles:
I think all the major scripting is done for this feature.  It can be merged in anytime.

DedrickEnc:
I am at Tshikaji trying to clean data and fix some problems, I suggest to merge it when back to kinshasa so it will be considered as one of major enhancement of the new bhima release.

jniles:
One thing to note that I just realized:  If we ever plan on using [cluster](), two things have to be kept in mind: 
 1. The PluginManager should only be defined once, in the master node
 2. We will have to figure out a new solution for HTTP requests made to plugins (the request might go to a child server that does not have access to the PluginManager.

Naturally, we can cross these issues when we come to them, but it is important to realize that having a cluster of node servers will each spin up (or try to) a cluster of plugins.

server/app.js
```diff
@@ -25,6 +25,9 @@ require('./config/express')(app, authenticate);
 // Link routes
 require('./config/routes').initialise(app);
 
+// Load and configure plugins
+require('../plugins/pluginManager')(app, config);
```
DedrickEnc: The plugin manager exposed function is called with a full configuration object, I should send only the plugin property as long as you are using only plugin property at the plugin manager .

plugins/pluginManager.js
```diff
@@ -0,0 +1,167 @@
...
+
+
+/** wraps process.kill() */
+Plugin.prototype.kill = function (code) {
+  this.process.kill(code);
+};
+
+
+/** wraps process.on() */
+Plugin.prototype.register = function (event, callback) {
```
DedrickEnc: This function seems important but unused

jniles: It is being used in a separate branch of mine (plugins/event-monitor), which hasn't been pushed to GitHub yet.  It's part of a feature to monitor the status of plugins from the bhima web application over HTTP.

server/app.js
jniles: I agree and will clean this up soon.  We do need some way that the plugins have access to the db connector, application name, and other environmental variables, but that could be defined in their own configuration files.

plugins/pluginManager.js
```diff
@@ -0,0 +1,167 @@
...
+
+  _onStartup();
+}
+
+
+/** parses events and routes them to the correct plugin  */
+PluginManager.prototype.routeEvent = function (event, data) {
+
+  // parse the plugin name from the event
+  var params = event.split('::'),
```
DedrickEnc: Nothing Guarantee that the array will have two elements because the "::" mark is not a requirement for the API.

jniles: See comment on [line 154](https://github.com/IMA-WorldHealth/bhima/pull/554/files#diff-b94338ae957cdd5c7365874473305f41R154).  The API specifies that there should be a '::', but the code doesn't enforce it.  I'll put a try/catch block.

PR #564, opened 2015-06-22
Improve unit Budget, adding some account of class 1 and 2 used in budget

lomamech:
Improve unit Budget, adding some account of class 1 and 2 used in budget, and sanitaze code


client/src/js/directives/findaccount.js
```diff
@@ -1,120 +1,117 @@
...
-	if (selectedAccount) {
-	  return selectedAccount.account_txt + ' [' + selectedAccount.account_number + ']';
-	  }
-	else {
-	  return account ? account.account_txt : '';
-	}
+				if (account) {
+	  			if (!isNaN(account)) {
+	    			var data = getAccount(scope.accountId);
+            console.log('dataHHHHHHHHHHHHHHHH',data);
```
jniles: Can we remove this?

client/src/partials/budget/budget.html
```diff
@@ -97,8 +97,8 @@
               <td class="budget-cash" colspan="3"><b>{{ "BUDGET.DISPLAY.TOTALS" | translate }}</b></td>
               <td class="budget-cash"><b>{{ totalBudget }}</b></td>
               <td class="budget-cash"><b>{{ totalBalance }}</b></td>
-              <td class="budget-cash"><b>{{ totalBudget }}</b></td>
-              <td class="budget-cash"><b>{{ totalBalance }}</b></td>              
+              <td class="budget-cash"><b>{{ totalSurplus }}</b></td>
+              <td class="budget-cash"><b>{{ totalDeficit }}</b></td>
```
jniles: It is much better practice to namespace these.  Something like:
```html
<td>{{ totals.surplus }}</td>
<td>{{ totals.deficit }}></td>
```

client/src/partials/budget/edit/edit_budget.html
```diff
@@ -3,160 +3,150 @@
...
+                     ng-disabled="session.totalBudget <= 0.0"
+                     ng-change="recompute()"
+                     ng-model="session.autoAdjust"></b>
+                  </td>
+                  <td style="vertical-align: middle">
+                    <span class="pull-right">{{ "BUDGET.EDIT.TOTAL" | translate }}</span>
+                  </td>
+                  <td style="vertical-align: middle">
+                    <input type="number"
+                     ng-class="{'form-bhima-red-border' : session.totalBudget <= 0}"
```
jniles: There are other ways to detect and handle errors in forms using a controller and a `formController.input.$error`-type properties.  We should research this in the future so that we don't re-implement the same logic in each controller.

server/controllers/uncategorised.js
```diff
@@ -248,7 +276,9 @@ exports.listInExAccounts = function (req, res, next) {
 
   function process(accounts) {
```
jniles: Are 1_2 accounts really income/exchange accounts?  I thought they were capital accounts.  Why are they in the filter for `exports.listInExAccounts`?

lomamech: I also just add accounts Class 5, Patrick accounting consultant says
For some account for Class 1
   The loans can be budgeted, because they are considered a corporate asset that can not happen in an account  
   product (Class 7) but rather in the 16 
For some account for Class 2
  The acquisition of the assets may be budgeted because the accounting point of view this is not an expense 
   account that is used, rather an account of class 2

For some account for Class 5
  because of the accounting perspective, customer payments can be made by check or by payment order.
  This represents a product for the company.

jniles: Your logic makes sense.  However, the route is named `exports.listInExAccounts`, which is confusing because it queries more than just income/expense accounts.  In the future, please make sure your function names match what they actually do.

PR #578, opened 2015-07-09
Adding ID_Transaction in receipt of Donation and Integration during C…

lomamech:
Adding ID_Transaction in receipt of Donation and Integration during Confirmation in stock, and adding Unit Report of Integration

### DIFF
--- a/client/src/partials/reports/stock_integration/stock_integration.js
+++ b/client/src/partials/reports/stock_integration/stock_integration.js
@@ -1,16 +1,13 @@
 angular.module('bhima.controllers')
-.controller('stock_integration', [
+.controller('stockIntegration', [
   '$scope',
-  '$timeout',
-  '$translate',
+  '$window',
   'connect',
   'appstate',
   'validate',
-  'exchange',
-  function ($scope, $timeout, $translate, connect, appstate, validate, exchange) {
+  function ($scope, $window, connect, appstate, validate) {
     var session = $scope.session = {},
-      state = $scope.state,
-      allTypes = $scope.allTypes = '0,1';
+      state = $scope.state;
 
     $scope.selected = null;
 
@@ -135,7 +132,7 @@ angular.module('bhima.controllers')
     });
 
     $scope.print = function print() {
-      window.print();
+      $window.print();
     };
 
    function reconfigure () {
--- a/client/src/partials/reports/stock_integration/stock_integration.js
+++ b/client/src/partials/reports/stock_integration/stock_integration.js
@@ -1,0 +1,150 @@
+angular.module('bhima.controllers')
+.controller('stock_integration', [
+  '$scope',
+  '$timeout',
+  '$translate',
+  'connect',
+  'appstate',
+  'validate',
+  'exchange',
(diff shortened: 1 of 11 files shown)


client/src/partials/reports/stock_integration/stock_integration.js
```diff
@@ -0,0 +1,150 @@
+angular.module('bhima.controllers')
+.controller('stock_integration', [
+  '$scope',
+  '$timeout',
+  '$translate',
+  'connect',
+  'appstate',
+  'validate',
+  'exchange',
+  function ($scope, $timeout, $translate, connect, appstate, validate, exchange) {
```
jniles: `$timeout`, `$translate`, and `exchange` are not used in this controller.  Why import them?

server/controllers/report.js
```diff
@@ -870,18 +870,50 @@ function donation_confirmationRecords(params) {
...
 }
 
+function integration_stock(params) {
+  var p = querystring.parse(params);
+
+  var _start = sanitize.escape(util.toMysqlDate(new Date(p.start))),
+      _end =  sanitize.escape(util.toMysqlDate(new Date(p.end))),
+      _depot = sanitize.escape(p.depot);
+
+  var sql =
```
jniles: In ECMAScript 6 (the next version of JavaScript), the `\`` tick marks are used for [string templating](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/template_strings).  I know we have been using them in our SQL strings for a long time, but let's start avoiding them.  If you could remove them here, it would be a good start.

client/src/partials/reports/stock_integration/stock_integration.js
```diff
@@ -0,0 +1,150 @@
...
+        search($scope.options[0]);
+      });
+    });
+
+    appstate.register('enterprise', function (enterprise) {
+      $scope.enterprise = enterprise;
+    });
+
+    $scope.print = function print() {
+      window.print();
```
jniles: To call native functions, it is best to use [Angular's `$window` service](https://docs.angularjs.org/api/ng/service/$window).

client/src/partials/reports/stock_integration/stock_integration.js
```diff
@@ -0,0 +1,150 @@
+angular.module('bhima.controllers')
+.controller('stock_integration', [
```
jniles: `stockIntegration` is a better controller name (keeping with our mixedCase style)

client/src/partials/reports/stock_integration/stock_integration.js
```diff
@@ -0,0 +1,150 @@
...
+  '$timeout',
+  '$translate',
+  'connect',
+  'appstate',
+  'validate',
+  'exchange',
+  function ($scope, $timeout, $translate, connect, appstate, validate, exchange) {
+    var session = $scope.session = {},
+      state = $scope.state,
+      allTypes = $scope.allTypes = '0,1';
```
jniles: I may have missed something, but it doesn't look like you use this variable at all.  Why is it defined and why is it defined on the `$scope` object?

jniles:
Some edits to be made before this gets pulled in.

PR #591, opened 2015-08-05
Monthly balance

lomamech:
Update and Improve Monthly Balance Report


client/src/partials/accounts/create_account/create.html
```diff
@@ -218,7 +212,7 @@
                     <option value="">{{ "SELECT.REFERENCE"  | translate }}</option>
                   </select>
                 </div>
-                <div class="form-group">
+                <div class="form-group" ng-if="newAccount.type.type === 'balance'">
```
jniles: `newAccount.type.type`?

client/src/partials/accounts/create_account/create.html
```diff
@@ -265,18 +254,12 @@
                     </div>
                     <div class="radio">
                       <label>
-                        <input type="radio" name="option_is_asset" value="false" ng-model="editAccount.is_asset" >
+                        <input type="radio" name="option_is_asset" value="false" ng-model="editAccount.is_asset" ng-checked="editAccount.is_asset==0">
```
jniles: For consistency, it would be better to use triple equals `===`.

client/src/partials/reports/balance_mensuelle/balance_mensuelle.js
```diff
@@ -58,6 +108,8 @@ angular.module('bhima.controllers')
       session.sumOldCredit = $scope.balance_mensuelle.data.reduce(function sum(a, b) { return a + b.old_credit; },0);
       session.sumDebit = $scope.balance_mensuelle.data.reduce(function sum(a, b) { return a + b.debit; },0);
       session.sumCredit = $scope.balance_mensuelle.data.reduce(function sum(a, b) { return a + b.credit; },0);
+      session.solde_debit = $scope.balance_mensuelle.data.reduce(function sum(a, b) { return a + b.solde_debit; },0);
+      session.solde_credit = $scope.balance_mensuelle.data.reduce(function sum(a, b) { return a + b.solde_credit; },0);
```
jniles: There are six (!) loops through `$scope.balance_mensuelle.data` going on here. Here is a trick to reduce it to a single loop:
```js
var sums = {
  sumOldDebit : 0,
  sumOldCredit : 0,
  sumDebit : 0,
  sumCredit : 0,
  solde_debit : 0,
  solde_credit :0
};

sums = $scope.balance_mensuelle.data.reduce(function (summer, row) {
  summer.sumOldDebit += row.old_debit;
  summer.sumOldCredit += row.old_credit;
  summer.sumDebit += row.debit;
  summer.sumCredit += row.credit;
  summer.solde_debit += row.solde_debit;
  summer.solde_credit += row.solde_credit;
  return summer;
}, sums);

// Everything is now summed in one loop in `sums`!  You can attach `sums` to `session` as you want.
```

server/controllers/report.js
```diff
@@ -756,6 +756,69 @@ function stock_movement (params){
   return db.exec(requette);
 }
 
+function balance_mensuelle_all (params){
```
jniles: In SQL, we should stop use the '`' tick marks.  In JavaScript v6 they mean something different.

PR #595, opened 2015-08-07
Addition a fonction in more modules for research a patient

lomamech:
Addition a fonction in more modules for research a patient


jniles:
Thanks!

PR #596, opened 2015-08-07
Improve extra paiement

lomamech:
Improve module Extra paiement and rename this in Regularization paiement

### DIFF
--- a/client/src/partials/cash/extra_payment/extra_payment.html
+++ b/client/src/partials/cash/extra_payment/extra_payment.html
@@ -1,163 +1,212 @@
-<header data-header>
-  {{ "CASH.EXTRA.TITLE" |  translate }}
-</header>
-
-<nav>
-  <div class="pull-left">
-    <ol class="breadcrumb" style="float: right;">
-      <li><a href="#/"><span class="glyphicon glyphicon-home"></span></a></li>
-      <li class="active">{{ "CASH.EXTRA.TITLE" | translate }}</li>
-    </ol>
-  </div>
-
-  <div class="pull-right">
-    <button
-      style="margin-left: 5px;"
-      ng-click="reconfigure()"
-      class="btn btn-sm btn-default">
-      {{'PRIMARY_CASH.EXPENSE.SELECT_ACCOUNT' | translate}}
-      <span class="glyphicon glyphicon-repeat"></span>
-    </button>
-  </div>
-</nav>
-
-<main>
-  <div ng-if="!session.configured" class="row margin-top-10">
-    <div class="col-xs-12">
-      <div class="panel panel-default">
-        <div class="panel-heading">
-          {{ "PRIMARY_CASH.EXPENSE.SELECT_ACCOUNT" | translate }}:
-          <b>{{ "CASH.EXTRA.WAIT_ACCOUNT" | translate }}</b>
-        </div>
-
-        <div class="panel-body">
-          <div class="configure">
-            <span
-              style="font-size: 70px;"
-              class="glyphicon glyphicon-inbox">
-            </span>
+<div ng-switch="state">
+  <div ng-switch-default>
+    <main>
+      <div class="col-xs-12">
+        <div class="panel panel-default">
+          <div class="panel-heading">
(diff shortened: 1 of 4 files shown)


client/src/partials/cash/extra_payment/extra_payment.js
```diff
@@ -49,66 +46,125 @@ angular.module('bhima.controllers')
...
+          where : ['patient.debitor_uuid=' + id]
+        }
+      };
+      validate.process(dependencies, ['sales'])
+      .then(function (model) {
+        var sales = model.sales.data;
+        $scope.saleData = [];
+        sales.forEach(function (sale) {
+          $http.get('/ledgers/debitor/'+sale.debitor_uuid)
+          .success(function (rows) {
```
jniles: This generates an $http request for each debtor.  Is it possible to change the code so that we only generate one?

client/src/partials/cash/extra_payment/extra_payment.js
```diff
@@ -49,66 +46,125 @@ angular.module('bhima.controllers')
...
+            $scope.saleData = $scope.saleData.concat(items);
+            $scope.saleDataUnique = filterUnique($scope.saleData, 'transaction');
           });
-          $scope.saleData = $scope.saleData.concat(items);
-          $scope.saleDataUnique = filterUnique($scope.saleData, 'transaction');
-        });
+        });  
       });
+
+      function filterUnique(array, key) {
```
jniles: Here is a quick unique filter using reduce:
```js
var array = [1,2,2,3,4,1,5,3,1,6];
var uniqueValues = array.reduce(function (aggregate, item) {
  if (aggregate.indexOf(item) === -1) { aggregate.push(item); }
  return aggregate;
}, []);

console.log(uniqueValues); // [ 1, 2, 3, 4, 5, 6 ]
```

client/src/partials/cash/extra_payment/extra_payment.html
```diff
@@ -1,163 +1,212 @@
...
-  </div>
+          <div class="panel-body">
+            
+            <form name="configForm" novalidate>   
+              <div class="form-group">
+                <label class="control-label col-xs-4 required">
+                  {{ "SELECT.PATIENT" | translate }}
+                </label>
+                <div>
+                  <div class="pull-left form-group input-group">
```
jniles: Who does this have both a form-group and input-group class on it?

client/src/partials/cash/extra_payment/extra_payment.html
```diff
@@ -1,163 +1,212 @@
...
-              >
-            <span class="input-group-btn">
-              <button ng-click="setConfiguration(session.ac)" class="btn btn-default btn-sm square">{{ "FORM.SUBMIT" | translate }}</button>
-            </span>
+            </form>
           </div>
         </div>
       </div>
-    </div>
+      <script type="text/ng-template" id="template.html">
```
jniles: It would be good to make this template an html template somewhere.  What typeahead uses it?

client/src/partials/cash/extra_payment/extra_payment.html
```diff
@@ -1,163 +1,212 @@
...
+      </div>
+    </main>
 
-<script type="text/ng-template" id="AccountList.html">
-  <a>
-    <span bind-html-unsafe="match.label | typeaheadHighlight:query"></span>
-    <span><b>[{{ match.model.account_number }}]</b></span>
-  </a>
-</script>
+    <script type="text/ng-template" id="CreditorList.html">
```
jniles: It would be good to move this template into an html file.  Like "CreditorList.html".

client/src/partials/cash/extra_payment/extra_payment.js
jniles: Okay, so this case is more complex.  Here is a combined map/reduce that will perform what @mbayopanda was trying to do:
```
// filters and array of objects, returning a unique key/value
var uniqueValues = array.map(function (row) {
  return row[key];
}).reduce(function (aggregate, row) {
  if (aggregate.indexOf(item) === -1) { aggregate.push(item); }
  return aggregate;
}, []);
```

PR #597, opened 2015-08-07
Monthly balance

lomamech:
Improve code reduce six loop an a single loop, for the calculation of Sums


jniles:
Very nice, thank you.

PR #603, opened 2015-08-10
Research patient middle name

lomamech:
Adding element MiddleName during filtering patient for the research and improving sql request

### DIFF
--- a/server/controllers/patient.js
+++ b/server/controllers/patient.js
@@ -17,7 +17,7 @@ exports.searchUuid = function (req, res, next) {
 
   // compose the sql query
   sql =
-    'SELECT p.uuid, p.project_id, p.debitor_uuid, p.first_name, p.last_name, ' +
+    'SELECT p.uuid, p.project_id, p.debitor_uuid, p.first_name, p.last_name, p.middle_name, ' +
       'p.sex, p.dob, p.origin_location_id, p.reference, proj.abbr, d.text, ' +
       'dg.account_id, dg.price_list_uuid, dg.is_convention, dg.locked ' +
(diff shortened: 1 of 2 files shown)


server/controllers/patient.js
```diff
@@ -96,10 +96,15 @@ exports.searchFuzzy = function (req, res, next) {
     'WHERE LEFT(LOWER(CONCAT(p.last_name, \' \', p.first_name )), CHAR_LENGTH(?)) = ? OR ' +
       'SOUNDEX(p.first_name) = SOUNDEX(?) OR ' +
       'SOUNDEX(p.last_name) = SOUNDEX(?) OR ' +
-      'SOUNDEX(CONCAT(p.last_name, \' \', p.first_name)) = SOUNDEX(?) ' +
+      'SOUNDEX(p.middle_name) = SOUNDEX(?) OR ' +
+      'SOUNDEX(CONCAT(p.last_name, \' \', p.first_name)) = SOUNDEX(?) OR ' +
+      'SOUNDEX(CONCAT(p.first_name, \' \', p.last_name)) = SOUNDEX(?) OR ' +
+      'SOUNDEX(CONCAT(p.first_name, \' \', p.last_name, \' \', p.middle_name)) = SOUNDEX(?) OR ' +
+      'SOUNDEX(CONCAT(p.last_name, \' \', p.middle_name, \' \', p.first_name)) = SOUNDEX(?) OR ' +
```
jniles: The way SOUNDEX works, you don't need to have lines 102 and 103.  They will product e the same as lines 100 and 101 for shorter inputs.  I think this is doing too much SQL work, and may take a long time when we have many patients...

PR #604, opened 2015-08-10
Pdf report

DedrickEnc:
In this PR the pdf report are mae for : bilan and compte de resultat.


client/src/partials/reports_proposed/result_account/result_account.js
```diff
@@ -0,0 +1,109 @@
...
+      .success(function (result) {
+
+        // Expose generated document path to template
+        session.building = false;
+        $scope.generatedDocumentPath = result;
+      })
+      .error(function (code) {
+        session.building = false;
+
+        // TODO Handle error
```
jniles: Can we implement error handling for this?  How difficult would it be?

client/src/partials/reports_proposed/result_account/result_account.js
```diff
@@ -0,0 +1,109 @@
...
+  // Prototype document building module, requests document given configuration obejct
+  function ($scope, $http, $routeParams, $translate, $sce, validate) {
+
+    // Configuration objects optionally passed to /report/build - drives configuration UI
+    var session = $scope.session = {},
+        dependencies = {},
+        generatedDocumentPath = null,
+        serverUtilityPath = '/report/build/result_account';
+
+    var configuration = {
```
jniles: If these configurations are shared between reports, it might be useful to make a `ReportConfigService` that has these values (and others) so that each developer does not have to re-write them.

server/controllers/reports_proposed/data/account_result.js
```diff
@@ -0,0 +1,239 @@
...
+
+// expose the http route
+exports.compile = function (options) {
+  'use strict';
+
+  var deferred = q.defer(), context = {}, infos = {}, chargeData = {}, profitData = {};
+  context.reportDate = resultAccountDate.toDateString();
+
+  var sql =
+    'SELECT `acc`.`id` AS `accountId`, `acc`.`account_txt` AS `accounTxt`, `acc`.`account_number` AS `accountNumber`, ' +
```
jniles: It would be best not to use the "`" tick marks anymore.  It makes the code difficult to read.

server/controllers/reports_proposed/data/account_result.js
```diff
@@ -0,0 +1,239 @@
...
+        if(item.referenceId === reference.referenceId){
+          somDebit+=item.generalLegderDebit;
+          somCredit+=item.generalLegderCredit;
+        }
+      });
+
+      return somDebit - somCredit;
+    }
+
+    function getProfitNet (reference, currents){
```
jniles: You can make this function simpler:
```js
function getNetProfit(reference, currents) {
  return currents.reduce(function (sum, item) {
     return sum + (item.referenceId === reference.referenceId ? item.generalLedgerCredit - item.generalLedgerDebit : 0);
  }, 0);
}
```
In fact, it would be the most simple if you passed in `referenceId` rather than the `reference` object (saves a property lookup).

server/controllers/reports_proposed/data/account_result.js
```diff
@@ -0,0 +1,239 @@
...
+              net : 0,
+              previousNet : 0
+            });
+          }
+        }
+      }
+      return references;
+    }
+
+    function getChargeNet (reference, currents){
```
jniles: I see you use this format multiple times.  If you want to be really fancy, you can wrap this in a function that takes in a difference function.  For example:
```js
function getNetSum(referenceId, currents, fn) {
  return currents.reduce(function (sum, item) {
     return sum + (item.referenceId === referenceId ? fn(item) : 0);
  }, 0);
}

// now you can call
netCharge = getNetSum(id, currents, function (i) { return i.generalLedgerCredit - i.generalLedgerDebit; });
netProfit = getNetSum(id, currents, function (i) { return i.generalLedgerDebit - i.generalLedgerCredit; });
```

server/controllers/reports_proposed/data/account_result.js
```diff
@@ -0,0 +1,239 @@
...
+    profitData.previousAccountDetails = infos.previousAccountDetails.filter(function (item){
+      return item.sectionResultIsCharge === 0;
+    });
+
+
+    context.chargeSide = processCharge(chargeData);
+    context.profitSide = processProfit(profitData);
+
+
+    function processCharge (tbl){
```
jniles: The functions `processCharge` and `processProfit` are almost identical.  Is there anyway to make the code into a single function?  It will be easier to maintain in the long run.

PR #617, opened 2015-08-17
Update fiscal year lock

lomamech:
Update lock fiscal year, integration accounts of class 8 during locking of a fiscal year


client/src/partials/fiscal/fiscal.lock.js
```diff
@@ -85,10 +85,38 @@ angular.module('bhima.controllers')
...
+          var accountIndex = accountNumber.substring(0,2);
+
+          return (accountIndex === '81' || accountIndex === '83' || accountIndex === '85' || accountIndex === '89');
+        });
+
+        var otherProfit = session.array8.filter(function (item) {
+          var accountNumber = String(item.account_number);
+          var accountIndex = accountNumber.substring(0,2);
+
+          return (accountIndex === '82' || accountIndex === '84' || accountIndex === '86' || accountIndex === '88');
```
DedrickEnc: Using of modulo will be better, you convert the two first number into Number and you test if x % 2 is 0 (even number) or not (odd number)

jniles: In my opinion, we should __never__ parse the `account_number` field for bhima's logic.  Is there any other column that we have, or could add, that would let you identify the correct accounts?

lomamech:
I close this pull Request, i will do another pull request

client/src/partials/fiscal/fiscal.lock.js
DedrickEnc: Why we should not?

PR #621, opened 2015-08-21
Fiscal year update

lomamech:
Update the process for closing a fiscal Year


client/src/partials/accounts/create_account/create.js
```diff
@@ -135,6 +144,11 @@ angular.module('bhima.controllers')
       session.accountClass = $scope.checkClass($scope.newAccount.number);
     }
 
+    function typeAccountClass () {
+      console.log('A la recherche de Type Account');
```
DedrickEnc: Le console log inutile peuvent etre effac2s

client/src/partials/accounts/create_account/create.js
```diff
@@ -177,13 +190,22 @@ angular.module('bhima.controllers')
...
+      if($scope.editAccount.is_asset === 'true' || $scope.editAccount.is_asset === "1"){
         $scope.editAccount.is_asset = 1;
-      } else if ($scope.editAccount.is_asset === 'false' || $scope.editAccount.is_asset === 0){
+      } else if ($scope.editAccount.is_asset === 'false' || $scope.editAccount.is_asset === "0"){
         $scope.editAccount.is_asset = 0;
       } else {
         $scope.editAccount.is_asset = null;
       }
+
+      if($scope.editAccount.is_charge === 'true' || $scope.editAccount.is_charge === "1"){
```
DedrickEnc: Pourquoi pas faire : 
if($scope.editAccount.is_charget){} ....

DedrickEnc: Pourquoi pas faire : 
if($scope.editAccount.is_charge){} ....

DedrickEnc: pourquoi pas simplement : if($scope.editAccount.is_charge) ... ?

server/models/development/update/synt.sql
```diff
@@ -15,7 +15,6 @@ VALUES (114, 'compte de resultat', 'TREE.COMPTE_RESULTAT', 'pour voir le rapport
 -- Date: 2015-08-04
 -- By: Chris LOMAME
 
-
```
DedrickEnc: Pourquoi tes requettes ne concerne que les compte ohada pendant qu'il y a aussi les comptes PCGC ?

client/src/partials/accounts/create_account/create.js
```diff
@@ -96,6 +96,14 @@ angular.module('bhima.controllers')
         account.is_asset = null;
       }
 
+      if(account.is_charge === 'true'){
+        account.is_charge = 1;
+      } else if (account.is_charge === 'false'){
+        account.is_charge= 0;
+      } else {
+        account.is_charge = null;
+      }
```
jniles: You can condense this into one line! 
```js
account.is_charge = (account.is_charge === 'true') ? 1 : (account.is_charge === 'false') ? 0 : null;
```

However, I recommend that we do not use the strings 'true' and 'false'.  If you are using a checkbox, you can set `ng-true-value=1` and `ng-false-values=0` on the page where these are defined.  If they are coming from the server, they should be stored as a TINYINT field in the database.

client/src/partials/accounts/create_account/create.js
```diff
@@ -135,6 +144,10 @@ angular.module('bhima.controllers')
       session.accountClass = $scope.checkClass($scope.newAccount.number);
```
jniles: `$scope.checkClass()` is never called from the html view.  You don't need to attach it to the $scope.  Just make it a `function checkClass()`.

client/src/partials/accounts/create_account/create.js
```diff
@@ -135,6 +144,10 @@ angular.module('bhima.controllers')
...
     }
 
+    function typeAccountClass () {
```
jniles: This function only does one operation.  Why do you need to create a new function for it?

server/models/development/update/synt.sql
lomamech: Parce que la norme qui a était utilisée ne concerne que les comptes OHADA,

PR #634, opened 2015-08-25
Refactor Journal.js into smaller submodules

### LINKED ISSUE #47: Posting Journal Server-Side Modularization
As we approach the point of re-designing permission and routes, an important redesign goal is should be to modularize the architecture of the server posting journal code.  Current statistics are:

File | Length
------ | ------
`journal.js` | 1825 

The `journal.js` file is already set up for easy modularization.  It exports a central router that calls a given function based on the url parameters.  I propose the following redesign:

__Folder Structure__

```bash
- routes
  - journal
    - core.js         # contains the router export
    - coreutils.js   # contains utility functions common to each route (getTransactionId(), validUser(), ...)
    - engines       # contains the posting logic for each route in an individual file.
        - purchase.js
        - sale.js
        - cash.js
        - caution.js
# ...
```

In this model, `core.js` contains the map from URLs => engine and passes the parameters to each function. Ideally, the code would be incredibly generic and use promises.  An idea looks like this:

```javascript
module.exports = function router (db /*, dependencies */) {
  var utils, map;
  
  utils = require('./coreutils');

  map = {
    'purchase' : require('./engines/purchase')(db, utils),
    'sale' : require('./engines/sale')(db, utils)
    // etc ...
  };

  return function () {
    var args, table;
    args = Array.prototype.slice.call(arguments);
    table = args.unshift();   // get the first argument
    return (!!map[table]) ? map[table](args) : q.reject('No route for table : ' + table);
  };
};
```
This model can then handle any engines added later.  The only change is to add a line into the map.

Each engine then is responsible to doing checks to see if the data is properly set up.  The engine should also contain a rollback function, in case of failure.

Opinions?  Thoughts?

jniles:
This PR will close #47.  However, there are still some items left to do.

 - [x] refactor sale route
 - [x] refactor sale discard route
 - [x] refactor cash payment route
 - [x] refactor cash refund route (used to be cash/discard)
 - [x] refactor primary cash route
 - [x] refactor primary cash income route
 - [x] refactor primary cash expense route
 - [x] refactor primary cash refund route
 - [x] refactor payroll route
 - [x] refactor salary payment route
 - [x] refactor credit note route
 - [x] refactor consumption loss route
 - [x] refactor donation route
 - [x] refactor purchase route
 - [x] refactor employee invoice route
 - [x] refactor employee payment route
 - [x] refactor transfer route
 - [x] refactor convention payment route
 - [x] refactor promise routes
 - [x] refactor advance route
 - [x] refactor confirm (?) route
 - [x] refactor fiscal year result (?) route
 - [x] refactor integration route
 - [x] refactor tax payment route
 - [x] refactor cotisation payment route
 - [x] reversing stock

### EDIT

All routes have been migrated.  Testing in progress...


server/controllers/journal/cash.js
```diff
@@ -0,0 +1,411 @@
+var q = require('q'),
```
DedrickEnc: You put cash and cash discard logic in the same controller but why not for sale and sale discard? and also caution should be in the cash controller like cash and cash discard according to your logic.

jniles: That's fine - I didn't have time to even touch sale discard.  You will notice that sale discard is not checked on the list.

jniles:
I got this one now.

jniles:
@sfount, @dedrickenc - The PR can be merged anytime.  I've tested as much as I know how to test.

There are two bugs unrelated to this PR that crash `#/group_invoice` and `#/support`.  I believe it was because of #653.  There is some pretty broken code in those modules, so I recommend addressing the issues with them in a PR independent of this one.

Once this lands, I'll be able to merge #670 and #690 .  Let me know if you find bugs!

sfount:
This is really great work Jon. It will help a **huge** amount with maintainability and future code updates. I will pull this request to make sure it works with my build this evening. I'm happy for this to be merged whenever.

PR #635, opened 2015-08-25
Report patient age in higher granularity in find-patient directive

jniles:


DedrickEnc:
Even a few description will be better.

jniles:
Updated.

PR #638, opened 2015-08-26
Improve translation

lomamech:
Improve translation file and translate label untranslate


jniles:
Thanks!  This is a really good (and needed) PR!

PR #640, opened 2015-08-26
Little fix for checkboxe dysfunctional during Update

lomamech:
Little fix for checkboxe dysfunctional during Update


client/src/partials/inventory/depot/depot.html
```diff
@@ -68,7 +68,7 @@
         <div class="form-group">
           <label class="control-label col-xs-4">{{ "COLUMNS.IS_WAREHOUSE" | translate }}</label>
           <div class="col-xs-7">
-            <input type="checkbox" ng-model="session.new.is_warehouse">
+            <input type="checkbox" ng-value=1 ng-model="session.new.is_warehouse">
```
jniles: Should this be `ng-true-value=1`?

PR #642, opened 2015-08-26
Use protractor for e2e tests

jniles:
This PR is sets up the initial framework for protractor testing, with an example e2e test and an example unit test.  There is lots of more work and research to be done.

I'm looking into either a Travis or SauceLabs integration for testing so that not all of our tests run locally.  More to follow.


jniles:
Closing this because it is a work in progress and will require a more comprehensive PR.

PR #649, opened 2015-08-28
Build client + server to /bin with gulp

jniles:
This PR introduces lots of breaking changes to the application.  In short, it breaks everything.

### Building
The gulp file has been completely rewritten and moved to the top of the directory hierarchy.  This allows gulp to build both the client and the server from the same location.  The gulp output folder is now `/bin`, which creates subfolders `/client`, `/server`, and `/plugins`. 

**NOTE** The `/plugins/` folder is deprecated and here for compatibility purposes.  It will be removed once we complete plugin migration to individual git repositories, and `pluginManager.js` will reside in `server/lib/pluginManager.js`.

The gulp methods have changed to `gulp build-server` (copies server into `/bin/server`) and `gulp build-client`, which runs a large number of tasks to ultimately copy the client into `/bin/client`.  `gulp build` runs both client and server.

### How do I run the application?
It's quite simple.  Just do this:
```sh
# in the top level of the application
gulp build
cd bin
node server/app.js
```

### Why was this necessary?
The specific application was e2e tests locally.  By unifying the builds we can also unify the tests and allow use to spin up unit tests for the client and server, and e2e tests.

### What else changed?
 - We now concat  `app.js` into `bhima.min.js`.  Our index.html only loads one app file now!
 - `index.html` refers to items as `vendor/` instead of `lib/`.  It makes more sense.
 - Gulp tasks clean their targets using [del](https://www.npmjs.com/package/del) before writing to the target.
 - Lots of things got deleted.  We will miss them, but they were no longer useful.

### What else could be done?
 - [ ] `gulp test-e2e` & `gulp test-unit` should run tests for client and server.
 - [ ] auto-magically enclose angular files in a IIFE


jniles:
@sfount, @DedrickEnc I encourage you guys to check this branch out in see if you are okay with the new way of building.  Only once I hear back from both of you will I consider merging this.

DedrickEnc:
You are right, got it!

sfount:
@jniles. I was able to build both the client and the server first time, no problems there - I lightly used the server and everything checks out.

* Build process cleaned up nicely - great idea to delete all distribution files before starting
* Quite a complex process (not deal breakingly so, just something to keep in my mind)
* I was able to clearly understand each build step following the comment documentation
* 'gulp build' takes 9.5s on my machine. Not sure if you can do anything with this information but FYI!
* +1562, -39,229, something must have been done right
* I was unable to run the e2e tests but don't have enough time to properly set up protractor at the moment - I can also see that that section of the gulpfile is coming soon.

jniles:
@sfount, thank you for the feedback!

e2e tests do not work at the moment -- they prompted the need for this PR.  Having gulp run in the top level is the first step to getting a proper e2e test.

The build process can be optimized for time.  Right now, everything (__read__ jshint) is running all the time, but we likely can do less actions, and make those actions parallelized.  My builds take about as long as yours.  Previously, they took 20 seconds earlier in the year!

Are we a go or no go for merging on your end @sfount?

sfount:
Understandable; everything works well for me. go for merging :+1:

jniles:
I'll be updating the checklist as those features become available.

PR #651, opened 2015-08-28
Minor report fix

### LINKED ISSUE #650: Report generation silently fails
1. If no valid wkhtmltopdf instance is found on the server BHIMA still responds with URL of the PDF resource that doesn't exist.

2. If no PDF is found, the server crashes and does not report the issue. 

*wkhtmltopdf* should also be a dependency for BHIMA (potentially through bower) however that is a separate architecture issue.

sfount:
Very small error handling fixes. Closes #650


jniles:
LGTM.

PR #654, opened 2015-08-31
Move plugin manager into server

jniles:
The plugin manager is a library of the server, not a plugin itself.  This PR reflects that by moving the plugin manager into  `/server/lib/`.


DedrickEnc:
LGTM

PR #658, opened 2015-08-31
Add `npm run app` to ease build process

jniles:
You can now use `npm run app` from the top level to build and deploy build.  It's easier than lots of `cd`s.


DedrickEnc:
Good Idea.

PR #661, opened 2015-08-31
Implement end to end test structure

jniles:
This PR has been long overdue.  It sets up the basic framework for e2e test with selenium, protractor,  and gulp.

#### Install protractor & selenium
```sh
npm install -g protractor
webdriver-manager update
```

You will also have to build the test database additions in `test.sql`.
```sh
mysql -u root -p bhima < server/models/development/test.sql
```

#### How to run:
You need to have protractor, gulp-protractor, and selenium installed.  If you haven't done so already, you may need to run npm install.  _It is important_ that you install protractor globally and use webdriver-manager to install selenium.  The `protractor.conf.js` file will tell you where it expects to find them by default.

Once everything is installed, in one terminal type `npm run app` to compile and run the server.  Open up a separate window and confirm that protractor runs on its own by typing `protractor`.  Next confirm that `gulp-protractor` runs properly by typing `gulp client-test-e2e`.  Congratulations!  You ran an end 2 end test.

Currently, the default e2e test simply logs in and navigates to the patient registration page.  The login process is automated so future tests can take advantage of this and just write tests for their particular modules.

Let me know if you have any trouble with this PR.


server/models/development/update/synt.sql
```diff
@@ -115,3 +115,12 @@ UPDATE  `bhima`.`transaction_type` SET  `service_txt` =  'import_automatique' WH
 UPDATE  `bhima`.`transaction_type` SET  `service_txt` =  'group_deb_invoice' WHERE  `transaction_type`.`id` =5;
 UPDATE  `bhima`.`transaction_type` SET  `service_txt` =  'stock_loss' WHERE  `transaction_type`.`id` =13;
 UPDATE  `bhima`.`transaction_type` SET  `service_txt` =  'reversing_stock' WHERE  `transaction_type`.`id` =28;
+
+-- Add a test user (id:39) to user table
```
DedrickEnc: I suggest to create a separate file like development/test/test.sql to put these instructions because, this synt file will be executed to modified the database stored in the production server.

jniles: Good idea.  I've moved these to a new file.

PR #662, opened 2015-08-31
Redirect to /login when a server session doesn't exist

jniles:
This PR fixes an auth bug that will not redirect the user to /login when a session has been terminated.  We still need to come up with a good UI for this, but that will require a more thorough study of the problem.

#### Edit
Some nicer changes have also been implemented:
 1. The placeholder fields have been translated into fr/en.
 2. Two types of error messages, one for missing fields, and a separate for server-sent error.
 3. If the user is missing a username/password, the form never sends an $http request.
 4. Code brought closer in line with the angular style guide.


jniles:
Just a heads up - I'm going to do a bit more work on this before it should be merged.

jniles:
This can be merged anytime now.

PR #670, opened 2015-09-01
Improve tree

lomamech:
Improving of the tree, adding the possibility of having sub-trees


jniles:
Is it possible to remove the column `has_children` from the unit table?  Since we have the column `parent`, I don't think we need it.

lomamech:
If we remove the has_children column should whenever Requette launch a
server to check if an element has sub elements, while with the Requette
this column is done only for items that have sub elements

2015-09-04 9:09 GMT+01:00 Jonathan Niles <notifications@github.com>:

jniles:
@lomamech , in #310 I remove the `has_children` column, but it was never merged.  You can look at the code there to get an idea of how to implement the tree without `has_children`.

There should only be 1 HTTP request to `/tree` which loads all the modules, then the rest can be done in JavaScript.

jniles:
These are some nice improvements.  I'll do some tests and resolve the merge conflicts, and if everything goes correctly, I'll merge in the code.

PR #671, opened 2015-09-01
use rimraf instead of del for windows

### LINKED ISSUE #669: gulp-del does not work on windows
Using gulp-del on windows (git bash version 1.8.3), node v0.10.31, gulp-del fails without error.

jniles:
Potentially fixes #669.  According to [this SO question](http://stackoverflow.com/questions/28885074/running-del-npm-dependency-from-package-json-script-on-windows), the `del` npm module might have errors on windows.  This PR uses `rimraf` instead.


DedrickEnc:
It eequires a npm module

PR #672, opened 2015-09-01
fix and sanitaze translation file and complete translation file

lomamech:
fix and sanitaze translation file and complete translation file


jniles:
Wow.  I didn't realize we were keeping the `en.json` and `fr.json` files in alphabetical order.  I'll start doing that next time!

PR #676, opened 2015-09-02
Add modal windows for modules requiring the exchange rate

lomamech:
Add modal windows for modules requiring the exchange rate


jniles:
Why don't you create a new HTML file called `exchangeRateModal.html` with the template it it?  Then we are not injecting so much HTML into each page.

client/src/partials/caution/caution.js
```diff
@@ -251,6 +251,35 @@ angular.module('bhima.controllers')
...
     }
 
+    function haltOnNoExchange () {
+      if (exchange.hasDailyRate()) { return; }
+
+      var instance = $modal.open({
+        templateUrl : 'noExchangeRate.html',
+        backdrop    : 'static',
+        keyboard    : false,
+        controller  : function ($scope, $modalInstance) {
```
jniles: You could make an `exchangeModalController` somewhere else and then use it here.  For example:
```js
angular.module('bhima.controllers').controller('ExchangeModalController', ['$scope', '$modalInstance', ExchangeModalController]);

function ExchangeModalController($scope, $modalInstance) {
  $scope.timestamp= new Date();
  $scope.close = $modalInstance.dismiss;
  $scope.setExchange = $modalInstance.close;
}
```

client/src/partials/caution/caution.js
```diff
@@ -251,6 +251,35 @@ angular.module('bhima.controllers')
...
+          $scope.timestamp= new Date();
+
+          $scope.close = function close () {
+            $modalInstance.dismiss();
+          };
+
+          $scope.setExchange = function setExchange () {
```
jniles: `$scope.setExchange()` should probably redirect the user to `#/exchange_rate` using `$location`.

client/src/partials/primary_cash/income/transfer/transfer.js
```diff
@@ -138,6 +141,23 @@ angular.module('bhima.controllers')
...
+
+      var instance = $modal.open({
+        templateUrl : 'partials/exchangeRateModal/exchangeRateModal.html',
+        backdrop    : 'static',
+        keyboard    : false,
+        controller  : 'exchangeRateModal'
+      });
+      
+      instance.result.then(function () {
+        $location.path('/exchange_rate');
```
jniles: Since all of the call $location.path('/exchange_rate') on exit, I would put that in the exchangeRateModalController.

These changes look great!

lomamech: Sorry, I have not found a better way to put $location.path('/exchange_rate')
in the controller

2015-09-03 22:09 GMT+01:00 Jonathan Niles <notifications@github.com>:

jniles:
@lomamech, thanks for trying!  Good work, I'll get this into development.

PR #677, opened 2015-09-02
Amending bilan pdf report

DedrickEnc:
This PR contains :  
- correction of a balance formula to calculate balance by taking in account the account side (active or passive)
- creation of a report service to expose common data and function even if there is little data to expose for now
- internationalization of bilan pdf
- and many minors fixing
I preferred to keep both processAsset and processPassive function for clarity because each function has many lines, so using a conditional structure to join them will make one unclear big function.

### DIFF
--- a/server/controllers/reports_proposed/data/bilan.js
+++ b/server/controllers/reports_proposed/data/bilan.js
@@ -15,11 +15,26 @@ exports.compile = function (options) {
   console.log(i18nBilan);
 
   var deferred = q.defer(), context = {}, infos = {}, assetData = {}, passiveData = {};
+
+  //populating context object
+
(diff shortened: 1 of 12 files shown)


server/controllers/reports_proposed/data/bilan.js
```diff
@@ -251,10 +287,10 @@ exports.compile = function (options) {
           somCredit+=(item.generalLegderCredit);
         }
       });
-      return somDebit - somCredit;
+      return isActif == 1 ? somDebit - somCredit : somCredit - somDebit;
```
jniles: Please use triple equals `===` to make JSHint happy.

jniles:
Tested.  LGTM.

PR #686, opened 2015-09-03
Improve payroll

lomamech:
- Improved multi payroll module to prevent requettes the server for cofigurations that does not have data
- Fix form for create Taxe


jniles:
This looks much better.  I think we probably should redesign the way these requests are structured (in the future), but this is a good hotfix.

PR #688, opened 2015-09-03
Localisation and reworking currency

sfount:
In this PR:

1. Localisation using ngLocale, configures angular filters based on locale - the primary change has been updating dates from only US format to respecting the user language. _(Note: this does not apply to the currency filter - as described in this pull request)__

2. Forked the native angular currency filter - it is essential that currencies in BHIMA are __not__ dependent on locale, they must respect the original transactions currency. This filter enables:
* Safe alternative to the current intlcurrency filter - no unsafe data binding 
* Specify BHIMA currency to request correct formatting 
*

### DIFF
--- a/client/src/i18n/currency/fc.json
+++ b/client/src/i18n/currency/fc.json
@@ -1,0 +1,29 @@
+{
+  "CURRENCY_SYM": "Fc",
+  "DECIMAL_SEP": ",",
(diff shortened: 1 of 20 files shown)


sfount:
Commits where duplicated attempting to resolve very minor merge conflicts through rebase && pull. (Retrospectively not my best idea ever). I will resolve this and create another issue.

Issue #663, opened 2015-08-31
Build a robust application structure


sfount:
BHIMA is currently about 50% responsive, many components and screens break or tear on smaller sized screens and not many of the application pages are consistent with how to do layout. 

As well as this the current overall application structure was a CSS prototype that worked and was kept around, elements like fixed header, nav and footer were then added on top. 

I think it would be beneficial from a support/ reliability perspective to carefully architect the overall application structure to account for:
* Both small and large screens 
* Easy to read and write syntax structure 
* __Flexible components that don't cause unexpected behaviors (currently if the programmer doesn't implement the header or the footer on an application page the page shrinks and is still offset by those non existent elements - from a developer perspective having elements that simply take up their space when they exist would be a much better guarantee when thinking about how to implement a certain page)__

Unexpected design errors are usually easy to work around (position : absolute hack for example) but I have certainly underestimated the amount of development time they take every time something new has to be designed or added to the application. 

This is a difficult decision as the software is currently being used and this would involve tweaking the UI. However I think the benefits gained from a reliable, easy to add to and evolve structure would far outweigh the cons. It's also worth noting that the longer this doesn't happen the more users we will have to migrate. 

My proposal to cover the points I have described in the issue are:
* Design a side bar that has a fixed width - the % width causes it to become unusable on a medium sized screen 
* Allow the side bar to have a compact icon view (for example 50px) that can be expanded to a full utility view (250px). This would allow partials to be rendered with much more space on a smaller screen whilst simultaneously not comprising the utility of the full sidebar view. _(Unfortunately this is **much** easier to do from the left hand side of the screen which may involve a change)_
* Create a flexible application structure that allows for an **optional** header, utility menu and footer, whilst always making space for a clean independently overflowed content area. All of this must be specified with the partial view. This content area also must respect the position of the side bar. 

It would be really great to hear feedback on these thoughts - I understand that this will involve uprooting a lot of currently working design patterns (although hopefully no partial code will change) but I fear that if it is left any longer it will never get done. Please let me know if you think this should be addressed. 

(If this issue does not describe anything clearly I could put together some mockups/ sketches to demonstrate the proposed change)

jniles:
I'm heart and soul behind this change.  You didn't say this explicitly but if this will happen at all _it needs to be while bhima is still beta_.  (re: migrating more users to a new UI).

[...]

Issue #569, opened 2015-06-30
Migrate Currency and Date formats to ngLocale


jniles:
I'm not sure if we just didn't research properly or if this is a new addition to AngularJS, but there exists a module for proper configuration of dates and currencies called [ngLocale](http://www.angularjshub.com/examples/localization/simplelocalization/).  [Here](https://raw.githubusercontent.com/angular/angular.js/master/src/ngLocale/angular-locale_fr-be.js) is the Belgium French locale, which we could easily adapt to using Congolese Francs and remove the "intlcurrency" filter, which requires unsafe HTML injection.

Sticking to angular's way will let us write awesome code like `{{ new Date() | date:"Full Date" }}` and `{{ results.total | currency }}` without fear.  We may have to modify the currency filter slightly to ensure that we can use USD and FC in the same page.

sfount:
I've been looking into this and it looks like it should be possible - there are a few complications as you mentioned re : multiple different currencies in one page but I'll throw together a prototype to make sure it works without being too obtrusive. 

FYR https://raw.githubusercontent.com/angular/angular.js/master/src/ngLocale/angular-locale_fr-cd.js they got this.

jniles:
Yeah, @DedrickEnc thought that locale's translations were quite humorous, and they aren't exactly the formats we want them in. But it seem easy enough to roll our own!

Awesome, Iook forward to your prototype.

sfount:
Quick progress review so far - one more decision we have to make and I will submit a pull request tomorrow. 

Using ngLocale for dates alone is definitely worth it, seeing the application now without a French date format doesn't look right, this is fairly easy to do and hasn't caused a problem. 
There are two options for serving a locale file, essentially a script that is run at the application startup that defines a number of formats and options within angular:
* Templating the correct locale file into the application HTML based on user preference and browser/system settings; this would require a server infrastructure change however this is something I think we should eventually decide to do
* Serve the locale file dynamically. I have done this because it is very similar to the way ng-translate works and respects patterns the application currently uses. 

The complication comes when displaying multiple different currencies on one page using the default currency filter - this is simply /not/ supported. ```{{amount | currency}}``` will display the format of the current locale. 

The solution I propose is to write a new currency filter that is essentially a fork of the native angular filter - it would use exactly the same locale settings and configuration but accept a setting to define which currency within BHIMA should be displayed. This would also mitigate the unsafe HTML binding issue. 

This then is the question:

[...]

PR #689, opened 2015-09-04
Localisation and reworking currency

sfount:
Attempt 2.

In this PR:

1. Localisation using ngLocale, configures angular filters based on locale - the primary change has been updating dates from only US format to respecting the user language. __(Note: this does not apply to the currency filter - as described in this pull request)__
2. Forked the native angular currency filter - it is essential that currencies in BHIMA are __not__ dependent on locale, they must respect the original transactions currency. This filter enables:
  * Safe alternative to intlcurrency, no html binding is used
  * Filter independently of current locale
  * Specifiy BHIMA currency ID to request formatting
  * Render strong warnings if a non-supported currency is used or a mistake is made 
  * Asynchronously filter based on available JSON configuration files
3. Updated the database to remove formatting from currency line and create relationship with format configuration files
4. Service to manage async fetching and requesting of currency format files. This can be used by any module that requires information about currency format information (selects showing available currency options) however is primarily intended to drive the currency filter. 
5. Deprecated the intlcurrency filter - please note this PR request does not replace all instances of the filter. For now it simple returns an error message on use. 

An example of how the currency filter can be used
```
{{ transaction.total | currency:transaction.currency }}
```

The currency filter provides the configuration boolean ```requiresCurrencyDefinition``` By default this is set to false and the enterprise currency is used if no currency is specified. If this flag is set the filter will display a clear error stating that the developer should explicitly define a currency id. The development team can decide if it is safe to allow implicit use of the enterprise currency.

If this change is accepted I am happy to merge the very small conflicts between this branch and ```development```. (See my attempt at being clever here #688).

### DIFF
--- a/client/src/js/app.js
+++ b/client/src/js/app.js
@@ -754,7 +754,7 @@
 
     // TODO Hardcoded default translation/ localisation
     tmhDynamicLocaleProvider.localeLocationPattern('/i18n/locale/angular-locale_{{locale}}.js');
-    tmhDynamicLocaleProvider.defaultLocale('fr-fr');
+    tmhDynamicLocaleProvider.defaultLocale('fr-be');
   }
 
   // Logs HTTP errors to the console, even if uncaught
--- a/client/src/js/app.js
+++ b/client/src/js/app.js
@@ -1,7 +1,7 @@
 (function (angular) {
   'use strict';
 
-  var bhima = angular.module('bhima', ['bhima.controllers', 'bhima.services', 'bhima.directives', 'bhima.filters', 'ngRoute', 'ui.bootstrap', 'pascalprecht.translate', 'LocalForageModule']);
+  var bhima = angular.module('bhima', ['bhima.controllers', 'bhima.services', 'bhima.directives', 'bhima.filters', 'ngRoute', 'ui.bootstrap', 'pascalprecht.translate', 'LocalForageModule', 'tmh.dynamicLocale']);
 
   function bhimaconfig($routeProvider) {
     //TODO: Dynamic routes loaded from unit database?
@@ -746,9 +746,15 @@
     });
 
     $translateProvider.useSanitizeValueStrategy('escape');
-
-    //TODO Try and assign the previous sessions language key here
+  
     $translateProvider.preferredLanguage('fr');
+  }
+
+  function localeConfig(tmhDynamicLocaleProvider) { 
+
+    // TODO Hardcoded default translation/ localisation
+    tmhDynamicLocaleProvider.localeLocationPattern('/i18n/locale/angular-locale_{{locale}}.js');
+    tmhDynamicLocaleProvider.defaultLocale('fr-fr');
   }
 
   // Logs HTTP errors to the console, even if uncaught
@@ -778,6 +784,7 @@
   // configuration
   bhima.config(['$routeProvider', bhimaconfig]);
   bhima.config(['$translateProvider', translateConfig]);
+  bhima.config(['tmhDynamicLocaleProvider', localeConfig]);
   bhima.config(['$httpProvider', authConfig]);
   bhima.config(['$localForageProvider', localForageConfig]);
   // run
--- a/client/src/js/filters/filters.js
+++ b/client/src/js/filters/filters.js
@@ -53,10 +53,12 @@
           console.log('patterns 1', formats.PATTERNS[1]);
           console.log('group sep', formats.GROUP_SEP);
           console.log('decimal sep', formats.DECIMAL_SEP);
-
-          formats.DECIMAL_SEP = currency.get(targetLocale).decimal;
-          currencySymbol = currency.get(targetLocale).symbol; 
-          
+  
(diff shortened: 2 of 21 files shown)


sfount:
Addresses #569

client/src/js/app.js
```diff
@@ -746,11 +746,17 @@
...
-    //TODO Try and assign the previous sessions language key here
+  
     $translateProvider.preferredLanguage('fr');
   }
 
+  function localeConfig(tmhDynamicLocaleProvider) { 
+
+    // TODO Hardcoded default translation/ localisation
+    tmhDynamicLocaleProvider.localeLocationPattern('/i18n/locale/angular-locale_{{locale}}.js');
+    tmhDynamicLocaleProvider.defaultLocale('fr-fr');
```
jniles: This is minor, but it might be best to bundle the [Belgian French](https://github.com/angular/angular.js/blob/master/src/ngLocale/angular-locale_fr-be.js) locale as default, rather than the France French.

client/src/js/filters/filters.js
```diff
@@ -1,109 +1,111 @@
...
+  .filter('boolean', function() {
+    return function (input) {
+      return Boolean(Number(input));
+    };
+  })
+
+  // Potential currency filter providing a specified locale 
+  // Ideally would leverage angular currency filter but this does NOT support
+  // passing through individual locale and doesn't expose utility methods 
+  .filter('intlcurrency', [
```
jniles: Is this filter still necessary?  In your [PR](https://github.com/IMA-WorldHealth/bhima/pull/689) you make no mention of it.  If not, can it be safely removed (all the modules have been migrated, or course)?

client/src/js/filters/filters.js
```diff
@@ -1,109 +1,111 @@
...
-          if (decimalDigits) {
-            value = value.slice(0, value.indexOf('.'));
-          }
+        // var templ = value.replace(/(\d)(?=(\d\d\d)+(?!\d))/g, '$1'+currency.get(id).separator);
+        // templ += '<span class="desc">' + currency.get(id).decimal + decimalDigits + '</span><span class="cur"> ' + currency.get(id).symbol +  '</span>';
 
-          var templ = value.replace(/(\d)(?=(\d\d\d)+(?!\d))/g, '$1'+currency.get(id).separator);
-          templ += '<span class="desc">' + currency.get(id).decimal + decimalDigits + '</span><span class="cur"> ' + currency.get(id).symbol +  '</span>';
+        // return $sce.trustAsHtml(templ);
+        return $sce.trustAsHtml('<span style="color : red;">Filter Depriciated</span>');
```
jniles: Welp.  I should have read more code.  Ignore [the comment above](https://github.com/IMA-WorldHealth/bhima/pull/689/files#r38726482).

client/src/partials/receipts/receipt.sale.js
```diff
@@ -1,13 +1,14 @@
 angular.module('bhima.controllers')
 .controller('receipt.sale', [
   '$scope',
+  '$locale',
```
jniles: Where is `$locale` used in the sale receipt?

jniles:
Minor fixes, but a merge with development.  On the whole, it looks awesome!  Can't wait to use this!

client/src/js/app.js
sfount: No problem with that. This does show a fairly lazy thing we do here using 'fr'/ 'fr-fr' strings as default.
This doesn't work if you don't have French for example. It's not the end of the world because it should just default back to angular's default (en-US) if it fails but the feature is clearly useful. 

My first naive thoughts are that you could template into the file/ index.html the enterprises default language and locale. This way it is ready to use during the angular bootstrap process. (app.js). (process mentioned brielfy here #663 )

client/src/js/filters/filters.js
sfount: Although it was an essay:
'5. Deprecated the intlcurrency filter - please note this PR request does not replace all instances of the filter. For now it simple returns an error message on use.'

This was mostly just a time issue. I grepped 'intlcurrency' and decided I would make it a separate commit!

client/src/js/app.js
jniles: True, true.  It was just an observation.

We should research a proper way to bootstrap the app.  The naive (probably easiest) way would be at build time with something like `gulp build --production --locales=en-US,fr-FR,fr-BE`.  I've heard some things about Angular2's router being about to lazy load routes, so maybe we can expect more support for bootstrapping dynamically within Angular.  Of course, we could try to bootstrap before Angular is even loaded, that idea is probably the worst in terms of clarity, maintainability, and portability.

client/src/js/filters/filters.js
jniles: Yup. My bad.  Overzealously posted before thoroughly reading. :S

sfount: Hey no worries, I appreciate how quickly you got around to it and the time that this kind of review takes.

PR #690, opened 2015-09-04
Fixing employee invoice

DedrickEnc:
This PR fixes the employee invoice interface by removing deprecated fucntion and using humaine readable information instead of uuid


jniles:
I'll merge this in after the journal refactor -- you did some work in the journal.js file which will be very painful to merge with the journal refactor branch.  It will be easier to merge these changes after the refactor has made it into `development`.

server/controllers/journal.js
```diff
@@ -822,7 +822,7 @@ function handleEmployeeInvoice (id, user_id, done) {
   var references = {}, cfg = {};
 
   function handleResult (results) {
-    if (results.length === 0) {
+    if (results.length <= 0) {
```
jniles: It doesn't make a lot of sense to check if `results.length <= 0`.  Why did you change it?

PR #695, opened 2015-09-04
Replace intlcurrency with currency

sfount:
Replaced all instances I could find of intlcurrency with the new currency filter. As this feature is used so much throughout the application this will allow it to be tested as soon as possible. 

**Please note I have not tested every page individually - whilst I was careful to respect a pages structure when swapping the filters it is essential that any issues are reported referencing this PR.** In the next few weeks we will be performing a number of changes that alter every single page in the application - it is worth designing an effective testing strategy before upgrading any installs.


jniles:
@sfount Can you resolve the merge conflicts before I begin testing?

sfount:
To my knowledge I couldn't do that without doing the merging process locally and then manually taking those changes and applying them to my local feature branch. 

Last night (?this morning) I tried using rebase and I thought I was a genius but it turns out quite the opposite. I did:
* Rebase development branch onto feature branch (resolving all conflicts) 
* Pull from origin (sfount) feature branch - pretty clever, even though it's behind
* Push to feature branch - now everything can be merged 

Of course you can see how well that went. It seems pretty obvious now that you get two of each commit; you're fetching and merging your commits on top of the forward rebased branch. 

Anyway - is there a secret I'm missing?

jniles:
I think this will do it.

```sh
git checkout depreciate_intlcurrency
git merge development
# fix conflicts
git commit -am "Pulled in development"
git push origin depreciate_intlciurrency
```

jniles:
`git rebase` only works if you have not opened a PR.  The problem is `git rebase` actively rewrites your history, which is fine if it's only _your_ history.  If you've shared it with the world (via a pull request), then you have to close the PR, rebase, and open a new one.

`git merge` will not touch the old history, but will instead create a new commit with the merged changes.  That is okay to share with the world since it is just an addition.

... at least.  That's my understanding.  There is some magic with `--no-ff` that changes what I said a little bit, somehow, but I haven't quite mastered that yet.

sfount:
Ah yes. But this will convolute my PR with the world and their pets commits right? Anything that's been added to development with show up here as I understand it. Maybe I've made that up.

jniles:
Not with merge.  It will be a single commit.  I don't believe the old changes will show up, since it will only compare what has changed between this branch and development.  If you've pull in development, they will be identical branches, except for the code you've written.

jniles:
The currency rendering I've tested looks good.  I don't think it was this PR, but Journal Voucher's currency select now reads "undefined" for all options..  But the currency renders when you select one!  That can be fixed either now or later.

For the original purpose of this PR, everything looks good to me.  It has my go ahead to merge.

sfount:
Hi Jon, great catch, thanks for looking into this. 

The reason for this is actually nothing to do with the filter. During the separation of database currency and format I had originally removed the ```symbol``` field - duplication of data, no relationship between JSON object and ```symbol``` field etc. 

However I realised it was used in many places (like this) that had nothing to do with actually rendering currency... So I removed all other columns but left symbol; in the future planning to encourage developers to use the JSON configuration files available to the currency filter to display currency filter and phase out ```db.currency.symbol```. 

Anyway during development I must have removed the symbol request in the journal voucher to make it work for my database. I've added that back in but do let me know if you think some other action should be taken.

jniles:
@sfount, that sounds like a good plan.  Looks good now.

sfount:
@DedrickEnc Have you had a chance to check this? You can merge it at any time or let me know if there are any problems! Thanks

Issue #660, opened 2015-08-31
Use bower for client side dependencies 


sfount:
We currently do not have a list of client side dependencies required to run BHIMA, instead the current repository has many different 'vendor/lib' files that have been committed as a part of the BHIMA code base. 

I suggest that we use the [bower](http://bower.io/) package manager to store these in exactly the same way we use npm for server side dependencies. 

Our configuration would look something like this 

*bower.json*
```
{ 
  "dependencies" : { 
    "jquery" : "2.1.4",
    "angular" : "x.x..x",
    "angular-translate" : "x.x.x",
    ...
    }
}
```

This would result in an installation process of 
```
npm install 
bower install 
gulp build
```

This would be an opportunity to clearly define development vs. production dependencies and remove code from the BHIMA repository that does not belong to BHIMA. 

I understand that there are pros and cons however. What do you think? If this gets the go ahead I'll go ahead and transition the vendor structure and provide instructions on how to install.

jniles:
I don't see anything wrong with using bower for bhima's dependencies.

There are two considerations:
 1. I remember manually editing angular-ui-bootstraps templates.  Can't remember why, but you might run into errors there ... just a heads up.
 2. I know bower can manage some things like CSS.  Are we just using this for larger libraries (angular-translate, angular, jquery, slick, etc...)?  Or will will we manage bootstrap with bower?

Interested in hearing your feedback,

sfount:
I remember that as well (we had some issues upgrading angular ui to a new version to make use of new features with our older bootstrap styles or something - modifications occured :sweat:), slick grid has some custom CSS too. 

If that is the case we can move any CSS/ JS that has been modified for BHIMA into bhima/client/src/, maybe something like bhima-angular-ui or if we want to over complicate it we could fork the original repositories and then use those as dependencies within BHIMA (probably not worth it).

I would say anything that isn't BHIMA specific becomes a dependency, that includes bootstrap etc. The benefit as far as I can see is that if a new version or dependency needs to included actual files within the repository don't need to be changed - just a line or two of JSON.

DedrickEnc:
I agree, we bower bhima will be more structured. npm for sever side and bower for client side, then gulp to build the entire application. it should be good if we could use one tool for managing the server and the client dependencies but that is okay.

sfount:
Thanks for the feedback Dedrick. It is possible to use npm for client side dependencies however this is not normally done and is not as supported as bower for client libraries in my experience. 

[...]

PR #696, opened 2015-09-05
Integrate bower package manager 

sfount:
Please ensure the local team can build this before merging it into the ```development``` branch.
Bower integration proposal

Cons: 
* Bower is an 'unopinionated' package manager. There is no way of cleanly specifying just the distribution folder so downloaded dependencies comes to 7.4MB. (For reference ```node_modules``` currently sits at around ~85MB)

Pros: 
* Could not previously demonstrate the client side dependencies for BHIMA, this provides a convenient index. 

```
"dependencies": {                                                                                                               
      "Chart.js": "~1.0.2",                                                                                                         
      "angular": "~1.4.5",                                                                                                          
      "angular-bootstrap": "~0.13.4",                                                                                               
      "angular-chart.js": "~0.8.1",                                                                                                 
      "angular-dynamic-locale": "~0.1.27",                                                                                          
      "angular-localforage": "~1.2.3",                                                                                              
      "angular-route": "~1.4.5",                                                                                                    
      "angular-sanitize": "~1.4.5",                                                                                                 
      "angular-translate": "~2.7.2",                                                                                                
      "angular-translate-loader-static-files": "~2.7.2",                                                                            
      "angular-translate-loader-url": "~2.7.2",                                                                                     
      "jquery": "~2.1.4",                                                                                                           
      "jquery-ui.min": "https://code.jquery.com/ui/1.11.4/jquery-ui.min.js",                                                        
      "jquery.event.drag-new": "*",                                                                                                 
      "slickgrid": "~2.1.0"                                                                                                         
    }
```

* Updating libraries requires no changing of core repository files - just a version number. 

* Standardised method of adding and removing dependencies for future contributors/ developers.

It is also worth noting the angular version has been updated to 1.4.5 once this is merged, I don't believe this breaks any APIs. As many aspects of the application should be verified as working before this is merged, many crucial components have been replaced with slightly newer versions (checked to not be breaking to my knowledge).

This pull request requires the bower tool to be installed, this can be done with:
```npm install -g bower```

Bower can then be used exactly like npm. 

1. ```npm install -g bower```
2. Checkout PR 
3. ```bower install```
4. ```gulp build```
5. ```npm run app```


sfount:
Implements results of discussion #660

sfount:
There is a difference in the slickgrid library version numbers used with this install that causes a very minor row heading rendering bug in slickgrid (grid must be scrolled to display edit row headings). 

As this works perfectly well for now - if this PR is merged this will be filed as a seperate issue (match slick grid with known functioning dependencies.)

jniles:
Just leaving a note - when this PR lands, we need to update our installation guide with the changes made here (bower) and the new build system.

sfount:
I didn't have time to review everything but reference [this commit](https://github.com/sfount/bhima/commit/1e9f0b794834ee775af0a1008ba9b0836c279b1cc)

jniles:
Bower works for me on my home network.  Things work fairly well throughout.  I'll give it another test at IMA, and try it on a windows machine, then we should be pretty secure.

Interestingly, there is another slickgrid bug - Abolishing groups works fine, but group by account does not work until you try to filter something.  It also throws some console errors about not finding `$scope.account.get(g.rows[0].account_number)`.

While it is best practice to _not_ pull this branch into the development with these bugs, I'm still in favor of merging the two.  More of us can help track down the bug(s) in development, plus it will get the team used to managing dependencies with bower.

sfount:
Thanks for testing this Jon, the most important element is making sure it works for development in Congo, something I'm confident in as npm has worked so far. 

I agree that it is a good idea to get these two merged however I also found another discrepancy in this slickgrid version (resizing columns breaks the grid). 

As our principles going forward will involve making sure only proven, clean, maintainable code will enter the development branch I would be hesitant to knowingly introduce these bugs.  

As we know all of these issues have root in one difference in dependency version and _not_ in BHIMA code, unfortunately I don't have time to chase it down right now, maybe I'll be able to find it this evening.

jniles:
@sfount, can you push this branch to a new IMA-WorldHealth branch?  We can work on this collaboratively.  When I get tired of working on posting journal, I'll try to debug it.

I'm closing this, and we'll open a new PR when we have an idea what is going on and fixed it in a new branch.

sfount:
Sure, that sounds like a good plan. It is worth noting that the issue you mentioned is also present in the ```development``` branch with the grid unable to sort by account unless you have filtered something.

PR #703, opened 2015-09-07
Fixed error in the formula of calculating Results

lomamech:
Fixed error in the formula of calculating Results


jniles:
Should this `currency` filter have a currency_id associated with it?

Issue #701, opened 2015-09-05
Use ng-options for <select> elements


jniles:
In quite a few places, we have code that looks like this:
```html
<select>
  <option ng-repeat="i in list" ng-value="" ng-selected="">{{ i }}</option>
</select>
```

[Here](https://github.com/IMA-WorldHealth/bhima/blob/master/client/src/partials/inventory/groups/groups.html#L101) [are](https://github.com/IMA-WorldHealth/bhima/blob/master/client/src/partials/reports_proposed/grand_livre/grand_livre.html#L31) [some](https://github.com/IMA-WorldHealth/bhima/blob/master/client/src/partials/creditor/group/creditor_group.html#L74) [places](https://github.com/IMA-WorldHealth/bhima/blob/master/client/src/partials/debtor/debtor_group.html#L94) [where]() [this pattern](https://github.com/IMA-WorldHealth/bhima/blob/master/client/src/partials/reports/operating_account/operating_account.html#L18) [is used](https://github.com/IMA-WorldHealth/bhima/blob/master/client/src/partials/reference/reference.html#L106).

Using `ng-repeat` on the `<select>` makes the ng-model bind a string, which requires extra work on the programmer's part to figure out what the model was actually supposed to be bound to.

We should really try to avoid using this pattern in the future.  As part of our refactor towards the end of the year, we should ensure that all these elements are changed out for `ng-option`s.

[According to the angular documentation](https://docs.angularjs.org/api/ng/directive/select):

> In many cases, ngRepeat can be used on <option> elements instead of ngOptions to achieve a similar result.  However, ngOptions provides some benefits, such as more flexibility in how the <select>'s model is assigned via the select as part of the comprehension expression, and additionally in reducing memory and increasing speed by not creating a new scope for each repeated instance.

sfount:
This is an excellent suggestion for general practice and I agree we should go back and improve this pattern where possible. (In practice ng-options is __significantly__ faster).

However in some cases ng-options does not provide the functionality required. It is impossible to disable individual option elements on a select using this directive. Sometimes it useful to demonstrate to a user that an option is available but cannot be selected/ is disabled. This pattern is also used to title selects in a number of places. 

#700 is a good example of ng-options not providing the required functionality.

sfount:
My previous comment can be ignored as of #706

jniles:
[...]

Issue #700, opened 2015-09-05
UI Freezing on All Transctions State report


jniles:
On the "All Transactions State" report (`#/reports/global_transaction/`), the UI noticably lags for ~ 7 seconds.  This is shown clearly in the chrome dev tools - while the DOM content is loaded in ~900ms, the actual page load (after templating) is ~6.91 seconds.  See below.

![ui-freeze](https://cloud.githubusercontent.com/assets/896472/9699342/2ba56224-53d7-11e5-8755-7d5a6f5a0554.png)

Most likely, our ng-repeat on accounts is taking far to much time.  We can try to speed this up by using [track by](https://docs.angularjs.org/api/ng/directive/ngRepeat) and [one time data-bindings](http://www.bennadel.com/blog/2759-exploring-one-time-bindings-in-angularjs-1-3.htm).  Someone should investigate these options to see if they would work.

If all else fails, we can look at different measures, such as using typeahead.

sfount:
I've run a few performance tests on this issue. 

You are correct in identifying the performance bottleneck as the repeating options - however it is nothing to with either how elements are tracked or data binding. 

I experimented using angular native [one time data binding](https://docs.angularjs.org/guide/expression#one-time-binding) to ensure it was nothing to do with multiple updates on each digest. 
```<option ng-repeat="account in ::accounts.data"></option>```
This still causes a significant time delay - as I understand it this is simply because angular is creating and inserting thousands of DOM elements _individually_. 

As far as I can tell there are a few options:
* Create a directive that takes the array of accounts, compiles them into one big element (the select) and inserts this into the DOM in one action. I've proven this works before with the 'reportGroupCompile' directive even though I believe that was never actually used. This still just provides a big select of accounts however it will process instantly. 

* Typeahead on accounts 

* Account search directive similar to the patient search directive - this could be used anywhere we must define an account and provide additional features. For example typeahead to lookup accounts you already know the numbers for and click a 'view all/ reference' button to be able to search through all the accounts (in a modal or something). 

Another point worth making here is the current code calls a formatting method in the template 
```{{formatAccount(acc)}}```
This is __not__ the reason for the initial slow UI however this method is called multiple times on every digest loop. This results in:

* Initial page load : **14976** calls to formatAccount
* Every time anything changes : **~5000** calls to formatAccount

[...]

PR #706, opened 2015-09-07
Demonstrating new ng-options tools

sfount:
Using ng-options is **significantly** faster than using an ng-repeat on an <options> tag as discussed #701 and #700. 

This PR demonstrates this taking DOM load time from ~7 seconds to < 1 second. 

It also demonstrates the new ng-options API that allows the developer to specify when to disable options using the 'disable when' syntax 

```<select ng-options="account.displayString disable when (account.disabled) for account in accounts.data" ng-model="session.selectedAccount"></select>```


jniles:
This is so cool.  7s -->  ~1s is an incredible performance increase!

PR #714, opened 2015-09-08
Removing appstate instead SessionService

lomamech:
Removing appstate instead SessionService


jniles:
LGTM.

jniles:
@lomamech, in the future, it would be good to make sure each commit has a unique message.  It will make it easier to find out where problems were introduced in the future.

PR #716, opened 2015-09-08
Budget Analysis Report

mbayopanda:
In this PR, we have added : 
- A module 'Budget/Budget Analysis' to compare current budget data to previous budget data
- Export budget data comparison to csv


client/src/js/app.js
```diff
@@ -15,10 +15,26 @@ function bhimaconfig($routeProvider) {
     controller: 'editAccountBudget',
     templateUrl: 'partials/budget/edit/edit_budget.html'
   })
+  .when('/budgeting/analysis', {
```
jniles: You added these controllers (analysis and create) twice.  Please remove the duplicates.

client/src/partials/budget/analysis/analysis_budget.js
```diff
@@ -0,0 +1,436 @@
...
+      // Insert the budget totals into the account data
+      $scope.accounts.data.forEach(function (acct) {
+        // Current budget
+        if (acct.id in totals) {
+          acct.budget = precision.round(totals[acct.id], 2);
+        } else {
+          if (acct.type === 'title') {
+            acct.budget = null;
+          } else {
+            acct.budget = 0; // No budget means 0 budget!
```
jniles: Good comments.

client/src/partials/budget/analysis/analysis_budget.js
```diff
@@ -0,0 +1,436 @@
...
+            account.surplus = 0;
+          } else {
+            account.deficit = 0;
+            account.surplus = 0;
+          }
+        }
+        totalSurplus += account.surplus;
+        totalDeficit += account.deficit;  
+
+        //TODO if parent.depth exists, increment and kill the loop (base case is ROOT_NODE)
```
jniles: Is this still a TODO?  It looks like you've done it.

client/src/partials/budget/analysis/analysis_budget.js
```diff
@@ -0,0 +1,436 @@
...
+      function handleFiscalYearBudget(fiscal_year_id, data) {
+        $scope.fiscalYearBudget[fiscal_year_id] = data;
+      }
+    }
+
+    function loadFiscalYears(models) {
+      angular.extend($scope, models);
+    }
+
+    appstate.register('enterprise', function (enterprise) {
```
jniles: You do not need to use appstate here.  We provide a new service, called SessionService, that can give you the current enterprise, user, and project without needing a callback.
```js
angular.module('bhima.controlllers')
.controller('ExampleController', ['$scope', 'SessionService', BudgetAnalysisController]);

function BudgetAnalysisController($scope, SessionService) {
  $scope.enterprise = SessionService.enterprise;  // $scope.enterprise.id, $scope.enterprise.currency_id, etc are defined!
  $scope.user = SessionService.user;   // $scope.user.id, $scope.user.username, etc.. are defined!
  $scope.project = SessionSerivice.project;  // $scope.project.id, $scope.project.abbr, etc ... are defined!
}
```
This should help clean your code up.

client/src/partials/budget/analysis/analysis_budget.js
```diff
@@ -0,0 +1,436 @@
...
+        var def = $q.defer();
+        session.selectedPreviousFY = session.previous_fiscal_years.filter(function(fy) {
+          return fy.checked;
+        });
+        def.resolve(session.selectedPreviousFY);
+        return def.promise;
+      }
+    }
+
+    function exportToCSV() {
```
jniles: This is a good beginning.  Could you make this into a service?  The service should take in an array of JSON objects and a configuration object, and return a downloadable CSV file.  For example, here is how I would use it:
```
var data = [{ id: 1, name : 'Jaquline'}, {id:2, name: 'Bombo'}, { id : 4, name  : 'Yvete' }];

// trigger a CSV download using a service
CSVService.download(data, { strict : true, headers: { id : 'ID', name : 'Patient\'s Last Name'}});
```

The CSVService should first check that each JSON object in the array has all the same properties/keys.  For example, it should accept `[{ id :1 }, { id: 2}]`, but reject `[{id : 1}, {id :3, name : 'Frank'}]` and `[{x : 1} , {y:2}]`.

If any values are null, they should be empty strings.  If you pass the config option `strict` to the CSVService, it should throw an error if you are missing values.

Finally, the config option `headers` should give an object that describes what the header row of the CSV should be.

client/src/partials/budget/analysis/analysis_budget.js
```diff
@@ -0,0 +1,436 @@
...
+          return acct.classe === 6 || acct.classe === '6';
+        });
+      } else if (accountClass === 7 || accountClass === income) {
+        // Class 7 accounts
+        $scope.accounts.data = model.filter(function (acct) {
+          return acct.classe === 7 || acct.classe === '7';
+        });
+      } 
+
+      return $q.when(true);
```
jniles: Why do you need this to return a promise?

fixing/README.md
```diff
@@ -0,0 +1,2 @@
+Fixations directory
```
jniles: What is this?

client/src/partials/budget/analysis/analysis_budget.js
```diff
@@ -0,0 +1,420 @@
...
+      $scope.fiscalYearBudget = {};
+      for(var fy in selectedPreviousFY) {
+        var budget = { 
+          id           : selectedPreviousFY[fy].id, 
+          data         : null, 
+          totalBudget  : 0, 
+          totalBalance : 0 
+        };
+        $scope.fiscalYearBudget[budget.id] = {};
+        getBudget(budget.id);
```
jniles: The function name `getBudget(id)` sounds like it should get _and return_ a budget.  In other words, I would expect you to use it like this:
```
var budget = getBudget(budget.id);
```

If the function does not return anything, it isn't really 'getting' anything, is it?  Perhaps it needs a better name?  Or maybe to be refactored into several functions, one that gets a budget, another that does performs some logic on that budget?

client/src/partials/budget/analysis/analysis_budget.js
```diff
@@ -0,0 +1,420 @@
...
+      $scope.total.budget = precision.round(totalBudget, 2);
+      $scope.total.balance = precision.round(totalBalance, 2);
+      // Previous Budget Data
+      $scope.totalPreviousBudget = {};
+      session.selectedPreviousFY.forEach(function (fy) {
+        $scope.totalPreviousBudget[fy.id] = Math.round(totalBudgetFY[fy.id], 2);
+      });
+    }
+
+    function parseAccountDepth(accountData, accountModel) {
```
jniles: This function does two things:
 1. Parses the account depth and assigns account.depth to equal the depth in the tree
 2. Totals the surplus/deficit based on the account classe

These are two vary different pieces of logic.  Do you think they might belong in two different functions?

client/src/partials/budget/analysis/analysis_budget.js
```diff
@@ -0,0 +1,420 @@
...
+          handleFiscalYearBudget(fiscal_year_id, data);
+        });
+      }
+
+      function handleFiscalYearBudget(fiscal_year_id, data) {
+        $scope.fiscalYearBudget[fiscal_year_id] = data;
+      }
+    }
+
+    function loadFiscalYears(models) {
```
jniles: This function is one line of code.  Does it really need to be a function?

client/src/partials/budget/analysis/analysis_budget.js
```diff
@@ -0,0 +1,420 @@
...
+          var prevBudget = a.previousBudget[fy.id] === null ? '' : a.previousBudget[fy.id];
+          previousBudget.push(prevBudget);
+        });
+
+        var row = [a.id, a.account_number, a.account_txt, budget, balance, a.surplus, a.deficit].concat(previousBudget, a.type);
+        fileData.data.push(row);
+      });
+
+      // Exportation
+      exportFile.csv(fileData, 'Budget Analysis Report');
```
jniles: This is a great beginning!

client/src/partials/budget/analysis/analysis_budget.js
```diff
@@ -0,0 +1,420 @@
...
+.controller('budget.analysis', [
+  '$q',
+  '$scope', 
+  '$window',
+  '$translate',
+  'validate',
+  'precision',
+  'messenger',
+  'SessionService',
+  'util',
```
jniles: I don't think you ever use util.  Can you remove it?

PR #718, opened 2015-09-09
Restaure  report of tax paiement and cotisation paiement, update table structure set default value to zero instead  null, improve cotisation creation

lomamech:
- Restaure  report of tax paiement and cotisation paiement, 
- update table structure set default value to zero instead  null, 
- improve cotisation creation


client/src/partials/reports/cotisation_payment/cotisation_payment.html
```diff
@@ -20,8 +20,24 @@
                     <span class="input-group-addon">
                       <span class="glyphicon glyphicon-folder-open"></span>
                     </span>  
-                    <select class="form-bhima" ng-model="session.period_id">
-                      <option ng-repeat="getPeriod in getPeriods.data" ng-value="getPeriod.id">{{getPeriod.label}} :: {{getPeriod.dateFrom | date }} - {{getPeriod.dateTo | date }}</option>
+                    <select class="form-bhima" required ng-click="selecCotisations()" ng-model="session.period_id">
```
jniles: @lomamech, could you use `ng-options` here?  It is much better for performance.  See #701 and #706 .

client/src/partials/reports/cotisation_payment/cotisation_payment.html
```diff
@@ -20,8 +20,24 @@
...
+              <div class="form-group">
+                <label class="control-label col-xs-4 required">
+                  {{ "SELECT.COTISATION" | translate }}
+                </label>
+                <div>
+                  <div class="pull-left form-group input-group">
+                    <span class="input-group-addon">
+                      <span class="glyphicon glyphicon-briefcase"></span>
+                    </span>  
+                    <select class="form-bhima" required ng-model="session.cotisation_id">
```
jniles: `ng-options` can be used here as well.

jniles:
@lomamech, this is a really long PR title.  In general, if you made many changes to a particular module, it might be better to say "Tax and Cotisation Improvements" and describe the in the PR body.  It's good to have all that information though!

client/src/partials/reports/taxes_payment/taxes_payment.html
```diff
@@ -20,8 +20,7 @@
                     <span class="input-group-addon">
                       <span class="glyphicon glyphicon-folder-open"></span>
                     </span>  
-                    <select class="form-bhima" required ng-click="selecTaxes()" ng-model="session.period_id">
-                      <option ng-repeat="getPeriod in getPeriods.data" ng-value="getPeriod.id">{{getPeriod.label}} :: {{getPeriod.dateFrom | date }} - {{getPeriod.dateTo | date }}</option>
+                    <select class="form-bhima" required ng-value="getPeriod.id" ng-click="selecTaxes()" ng-options="getPeriod.id as format(getPeriod) for getPeriod in getPeriods.data" ng-model="session.period_id">
```
jniles: I don't think you need ng-value here.  If you have an ng-model, then it should automatically bind your value to the ng-model.

Also, use ng-change="selectTaxes()", rather than "ng-click=selectTaxes()".  Since you are using a _select_ it is more meaningful than _click_ which is for buttons.

server/controllers/report.js
```diff
@@ -474,7 +475,7 @@ function cotisation_payment(params) {
     'JOIN paiement p ON e.id=p.employee_id ' +
     'JOIN cotisation_paiement z ON z.paiement_uuid=p.uuid ' +
     'JOIN cotisation t ON t.id=z.cotisation_id ' +
-    'WHERE p.paiement_period_id= ' + id +
+    'WHERE p.paiement_period_id= ' + id + ' AND z.cotisation_id=' + cotisation_id +
```
jniles: It would be better to use db.exec() for paramter escaping:

```js
'WHERE p.paiement_period_id = ? AND z.cotisation_id = ?;';

db.exec(sql, [id, cotisation_id])
```

PR #720, opened 2015-09-10
Translating emplouyee state pdf report

DedrickEnc:
The employee state is translated and used a service (reportConfigService) to share common data and behavior.


jniles:
LGTM.

PR #727, opened 2015-09-11
Fiscal year enhancement

mbayopanda:
In this PR, we add:
- Number of months
- Confirm dialog when we want locking fiscal year
- Cancel buton


client/src/partials/fiscal/fiscal.update.js
```diff
@@ -118,6 +119,7 @@ angular.module('bhima.controllers')
       session.saved = false;
       session.saveError = false;
       $scope.fiscal = angular.copy(editCache);
+      $route.reload();
```
DedrickEnc: This sub controller can not have the responsibility of reloading the all page, let use a service to handle the Fiscal Year state.

client/src/partials/fiscal/fiscal.js
```diff
@@ -44,10 +45,13 @@ angular.module('bhima.controllers')
 
     // select a fiscal year for lock
     function selectLock(id) {
-      $scope.selected = null; // Skip selected for update style
-      $scope.selectedToLock = id;
-      $scope.active = 'lock';
-      $scope.$broadcast('fiscal-year-selection-lock-change', id);
+      var choice = confirm($translate.instant('FISCAL_YEAR.CONFIRM_LOCKING'));
+      if (choice) {
+        $scope.selected = null; // Skip selected for update style
```
jniles: Good comments.

PR #730, opened 2015-09-12
Fix: use IMA logo as favicon

jniles:
This PR adds a critical feature, the IMA logo as a favicon.  We are well on our way towards excessive branding.


sfount:
As this is such a small change I will test this and merge. Once everyone has had a chance to pull this change I will close the pull request.

PR #737, opened 2015-09-14
Update supplier management

lomamech:
- Remove column location_id from table supplier
- Update dependencies who use column location_id from supplier


jniles:
Are we going to make address required?

lomamech:
I think I can do it just to address 1, address 2 can stay Optional

PR #739, opened 2015-09-15
FIX: purchase invoices render properly

jniles:
Purchase invoices were entirely broken due to an errant string.  This
string has been removed, as well as hard coded values in the purchase
template.  The following other changes have taken place:
 1. Controller abides by Angular Style Guide Syntax.
 2. Enterprise address uses <address> to be more semantic
 3. Added <abbr> for abbreviations and associated translations

### DIFF
--- a/client/src/i18n/en.json
+++ b/client/src/i18n/en.json
@@ -1071,6 +1071,7 @@
    "NEW"                      : "New Enterprise",
    "PHONE"                    : "Phone",
    "PO_BOX"                   : "PO Box",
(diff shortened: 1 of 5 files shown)


DedrickEnc:
Is this task assigned to you from projeqtor?

jniles:
@DedrickEnc , not specifically, but it was completely broken.  To test inventory management, I needed to create a purchase order.  The application crashed.  So I fixed it, in order that I could do the rest of my work.

PR #742, opened 2015-09-15
Using session service, and Correction depot suppresing

DedrickEnc:
According to the meeting, depot suppression was his corrected and removing deprecated connect interface.

### DIFF
--- a/client/src/partials/inventory/depot/depot.js
+++ b/client/src/partials/inventory/depot/depot.js
@@ -5,11 +5,12 @@ angular.module('bhima.controllers')
   'validate',
   'connect',
   'messenger',
-  'appstate',
+  'SessionService',
   'uuid',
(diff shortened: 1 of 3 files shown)


client/src/partials/inventory/depot/depot.js
```diff
@@ -26,17 +27,17 @@ angular.module('bhima.controllers')
...
-    appstate.register('enterprise', function (enterprise) {
-      $scope.enterprise = enterprise;
-      validate.process(dependencies)
+    validate.process(dependencies)
       .then(startup);
-    });
 
     $scope.delete = function (depot) {
-      connect.basicDelete('depot', depot.uuid, 'uuid')
+      connect.delete('depot','uuid', [depot.uuid])
```
jniles: I don't think you want brackets (`[`) here.

jniles: Did this ever get changed?

PR #745, opened 2015-09-16
Fixing erros, Aliasing, new service injector Syntax

DedrickEnc:
For more clarty I introduced the new syntax for service injector as discussed,
In this PR, I corrected the error in purchase order page also

### DIFF
--- a/client/src/partials/purchase/view/purchase_records.js
+++ b/client/src/partials/purchase/view/purchase_records.js
@@ -1,73 +1,73 @@
 
 var purchaseRecords = function ($q, connect) {
-  var that = this;
-    var session = that.session = { purchase_type: 'indirect'};
-    that.purchase_filter = {};
+  var vm = this;
+  var session = vm.session = { purchase_type: 'indirect'};
+  vm.purchase_filter = {};
 
-    function init() {
-      that.selected = null;
-      var promise = fetchRecords();
+  function init() {
+    vm.selected = null;
+    var promise = fetchRecords();
 
-      promise.then(function(model) {
-        that.indirect_purchase = model[1];
-        that.direct_purchase = model[0];
-      });
-    }
+    promise.then(function(model) {
+      vm.indirect_purchase = model[1];
+      vm.direct_purchase = model[0];
+    });
+  }
 
-    function fetchRecords() {
-      that.selected = {};
-      var indirect = {
-        'tables' : {
-          'purchase' : {
-            'columns' : ['uuid', 'reference', 'cost', 'discount', 'purchase_date', 'paid']
-          },
-          'creditor' : {
-            'columns' : ['text']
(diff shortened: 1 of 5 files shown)


client/src/partials/purchase/view/purchase_records.js
```diff
@@ -1,24 +1,21 @@
-angular.module('bhima.controllers')
-.controller('purchaseRecords', [
-  '$scope',
-  '$q',
-  'connect',
-  function ($scope, $q, connect) {
-    var session = $scope.session = { purchase_type: 'indirect'};
-    $scope.purchase_filter = {};
+
+var purchaseRecords = function ($q, connect) {
```
jniles: It is better to always have `angular.module()` as the first line in the controller.  [You can see an example here](https://github.com/johnpapa/angular-styleguide#separate-data-calls).  That way, you can easily find where Angular binds the controller and see what its name is.   Like this:
```js
.controller('PurchaseRecordsController', purchaseRecords);

purchaseRecords.$inject = [/* some injects */];

// this will be hoisted to the top of the controller by javascript!
function purchaseRecords(/* some names for the injects */) {
  // code ...
}
```

client/src/partials/purchase/view/purchase_records.js
```diff
@@ -1,24 +1,21 @@
...
+  var that = this;
```
jniles: For controllers, the variable name `vm` is recommended by the angular style guide (see [here](https://github.com/johnpapa/angular-styleguide#style-y032)).  `vm` stand for view-model (since the controller binds the view).  For services, the name `service` is recommended (so that you always know you are in a service).

We can choose a different name, but it would be nice to be consistent.  The [login controller](https://github.com/IMA-WorldHealth/bhima/blob/development/client/src/partials/auth/login.js#L18) already uses `vm` as the local variable binding `this`.

client/src/partials/purchase/view/purchase_records.js
DedrickEnc:
```
var MyController = function($scope, greeter) {
  // ...
}
MyController.$inject = ['$scope', 'greeter'];
someModule.controller('MyController', MyController);

```
I used an official angular guide for developper, to write this code.
and I think, it seems logic and natural for me. 
1. I create a simple function which is a kind of object
2. I inject services to this object
3. I make this object a controller.

client/src/partials/purchase/view/purchase_records.js
DedrickEnc: You are rigth, that is good but less meanning full for me me according to the angular context, I will use vm.

jniles:
LGTM.  Let me know when it is ready to merge.

DedrickEnc:
This PR can be merged any time.

PR #749, opened 2015-09-18
Fixing purchase order confirmation

### LINKED ISSUE #740: Confirm Purchase Orders is Broken
When trying to confirm a purchase order, the server breaks with an MySQL error:
```sh
Error: ER_PARSE_ERROR: You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near '?)' at line 1
```

Expected Result: Purchase order confirmed, possibly a receipt
Actual Result: Server breaks, goes into undefined states, sometimes client refuses to reconnect.

DedrickEnc:
This pull request closes #740


client/src/partials/purchase/confirm/confirm.html
```diff
@@ -24,10 +24,10 @@
 
           <div class="alert alert-info po-alert">
             <strong>{{ 'PURCHASE.PURCHASE_TYPE' | translate }} : </strong>
-            <input type="radio" id="indirectPO" ng-model="session.is_direct" ng-value="false" ng-click="resetSelected()">
+            <input type="radio" id="indirectPO" ng-model="purchaseConfirmCtrl.session.is_direct" ng-value="false" ng-click="purchaseConfirmCtrl.resetSelected()">
```
jniles: Because we are using Controller As syntax, we no longer need the object `session`.  Could you remove it?

DedrickEnc:
I made the change, this PR can be merge any time

jniles:
Thank you

jniles:
Considering how much code is different between the indirect and direct purchase orders, I wonder if it wouldn't be better to make two separate modules out of them.  It would be much easier to maintain, certainly.

PR #762, opened 2015-09-23
Fixing service distribution error

mbayopanda:
In this PR, we have fixed an error which occur when we distribute medecine to service


jniles:
LGTM

Issue #757, opened 2015-09-22
Confirm Stock Integration is Broken


jniles:
The page confirm stock integration is broken.  

Steps to reproduce:
 1.  Use  stock integration module to register stock
 2. Attempt to confirm stock integration normally

Expected Result: Stock is confirmed, navigate away from page
Actual Result: Errors thrown on the server and console, originating from journal.js files.

jniles:
This is fixed in #764.

Issue #753, opened 2015-09-21
Proposal: rework server configuration into fewer, top-level files


jniles:
Currently, bhima uses a whole directory `server/config/` to configure and run the server.  Many of these files are small and contain ~~bad ideas~~ outdated configuration parameters.  It looks like this:
```
server/
    app.js
    config/
         express.js
         routes.js
     environment/
         errors.json
         server.js
     keys/
         # tls keys
```

Here is my new proposal:
```
config.js         # config.js replaces server.js
server/
   app.js          # app.js absorbs express.js
   controllers/
        index.js  # index.js replaces routes.js
```
... and we completely delete errors.json.  It was a *Bad Idea (tm)* (I can say that, since I wrote it).

Currently, app.js loads `config/express.js` which configures loggers, authentication, the database, and sessions, and more.  I see no advantage to doing this in a separate file.  We should put those configurations directly in the app.js file.

Furthermore, I think we are migrating all of our routes into controllers/ or lib/ functions.  To reinforce that our routes belong in controllers, we can create a `controllers/index.js`, which does the same thing as routes.js currently does.  Using the name index.js is appropriate for nodejs, since it means that app.js only needs to do `require('./controllers')(app, cfg)` to load the file.

I think that should cover the majority of cases.  Most projects keep their configuration in one place, at the top level in config.js[on].  This proposal will make building and testing much easier.

There is still a question of what to do with TLS keys.  I would put them in the top level here, but I am not sure of a better solution.  We could research that in conjunction with #707.

What do you all think @IMA-WorldHealth/owners ?

sfount:
Hi @jniles. All of the issues from the last few days are just coming through for me, I'm on a train so I'll just leave very quick initial thoughts and come back and clean this up with a good connection.

* the reason for the environment folder was actually to do with your proposal in #752. It was originally not intended to be anything to do with 'errors' or 'server' but allow us to define a core config that would then be extended (combined) with a development.js or production.js or testing.js depending on something like NODE_ENV. I'm not sure I'd call this a 'bad idea' but it was never fully realised. Bringing this up in both of these issues should cover how we should progress on that.
* no big issues about doing all express configuration in app.js however it is configuration and if properly documented would allow very clear explanation of I smaller components. there is actually quite a lot of non obvious code in this file that would be scattered around the application vs. 'This sets up and configures the HTTP server, this sets up the routes, this sets up the plugins' etc. I don't feel very strongly on this one and it should be fine either way
[...]

Issue #752, opened 2015-09-21
Proposal: reorganize  server environmental variables


jniles:
I propose that we reorganize server configuration into a fashion that is easier to debug, test, and configure for either development or production.  There are several ways of doing this:
 - [using a dot env file](https://github.com/motdotla/dotenv)
 - [using environment.json + config.js](http://stackoverflow.com/questions/8332333/node-js-setting-up-environment-specific-configs-to-be-used-with-everyauth)
 - roll your own (env.json or something)

I'm not particular on the type of environmental file we use.  The benefit of having a development versus production build is it allows us to configure:
 - database names in production, development, testing (`bhima`, `bhima_test`, `bhima_{hospital_name}`, etc)
 - authentication privileges (production vs development vs testing)
 - log levels (debug, standard, etc)
 - ... lots more ...

I'm particularly interested in being able to switch database connection at startup to a test database for both e2e testing and API integration testing.  On an `npm pretest` hook, we can switch node's environment to testing to use a test database, suppress logging, etc, and ensure our tests run smoothly.

What do you think?

jniles:
Since no discussion has been generated, I'm closing this.  I've implemented a really naive environmental variable switch via `proces.argv` to switch between development and production environments.  If our use cases get any more complex, we should come up with a better design.

Issue #751, opened 2015-09-21
Proposal: Move all tests to a top level directory.


jniles:
I propose we move all tests, server and client, to a top level directory (`test/` or `spec/`).  This will separate concerns from production and development  -- if production doesn't want tests, there is no need to even build that directory.

It also will allow us to use tools like mocha + chai to run tests that integrate across client and server in an easier fashion.

What do you think?

jniles:
Given recent changes in testing, this no longer makes sense.  Client and server tests are largely decoupled, relying on protractor + webdriver for the client and mocha + chai for the server.  The chai assertion syntax is shared across both, but this is not enough reason to move tests into a top level directory.

PR #764, opened 2015-09-23
Full rewrite of Stock and Inventory Management

### LINKED ISSUE #624: Stock Dashboard generates too many HTTP requests
The stock dashboard performs a denial-of-service attack by generating roughly 1,400 HTTP requests at once to the server.  I believe the code responsible is [the following](https://github.com/IMA-WorldHealth/bhima/blob/master/client/src/partials/stock_dashboard/stock_dashboard.js#L262):
```js
      $http.get('/getDistinctInventories/')
      .success(function (inventories) {
        inventories.forEach(function (item) {
          var inventory = {};

          stockControler.getStock(item.inventory_uuid)
          .then(function (data) { inventory.stock = data;})
          .then(stockControler.getStockMin(item.inventory_uuid)
                .then(function (data) { inventory.stock_min = data; })
                .then(stockControler.getStockSecurity(item.inventory_uuid)
                      .then(function (data) { inventory.stock_security = data;})
                      .then(stockControler.getStockMax(item.inventory_uuid)
// ... etc
```

Anyhow, the page crashes.  The server lives on, but generates an `undefined` exception.

We should never generate recursive HTTP requests based on data in the database.  This leads to an unknown number of HTTP requests that we cannot control.  It is better practice to have a single HTTP request that performs as many database requests as it needs before returning the data in a large JSON.

jniles:
This pull request rewrites a lot of stock modules.  For details, see the commit messages.  Here are some of the highlights:
 1. Many of the `#/stock/.../:depotId/` routes have been migrated to `#/depot/:depotId/.../` routes.  One prominent exception is the `#/stock` route itself, which needs to be change to `#/depots/`.  However, I felt this change was too large for this PR.
 2. A robust inventory and depot HTTP API has been implemented to allow querying 
    1. Expiration dates
    2. Consumption rates (both average and total)
    3. Stock Levels
    4. Stock donations
    5. Stock Lead Times
    6. And more...

 Most API routes take in a start and end date parameter, allowing flexible reporting.
 3. Many routes removed from `uncategorised.js` to various inventory, depot, or stock controllers.

### Important Notes
 1. Because stock is in reference to a depot, the reports contained within the stock page are also in reference to that depot.  This means that distributions to patients, services, and others will only be shown if they happened at that depot.

### What should be done after this PR

 - This PR makes many of the uncategorised routes obsolete.  These, as well as obsolete translations, should be removed.
 - Stock Consumption Loss needs a complete redesign to be both compatible and bug-free.  There will be an issue about this shortly (**update** see #766)
 - Add API tests for all documented API routes using `chai-http`.  This requires that #751, #752, and  #753 be resolved before progress can be made on testing.
 - Implement as many stock reports + dashboards as possible to determine if new API features need to added, or old ones revised.
 - Fully implement CRUD for all depot distributions, migrating clientside logic to serverside logic.

### Bonus

If you are using `DateService` in your client controllers, please consider using `DateService.util.str()` to convert your dates to a string before passing them to the database.  This service does not yield the timezone errors that plague our dataset so far.

Closes #624, #757 and #698.


jniles:
Closing this while extra fixes go in.

jniles:
@DedrickEnc this is ready for a second round of testing.  *Note that distribution loss is broken in this PR*.

I would like to continue working on stock.  I think this PR fixes more bugs than it introduces, but there are still more issues to work on.  My current plan is to work on:
 1. #777 -- this will allow me to clean up the depot's posting journal hooks, and introduce better error handling.
 2. #766  -- after stock distribution loss is cleaned up, our depot/stock interfaces should be bug free.
 3. Continue migrating to `/depots/:depotId/{{module-name}}` url scheme for everything currently in `stock/`.
 4. Fully implement the stock min/max calculates we discussed together for the stock status report (it currently defaults to the `stock_min` and `stock_max` properties).
 5. #779 -- this should round out the inventory features that bhima plans to support for version 1.

What do you think?   Also interesting in hearing @sfount's opinion.

jniles:
@sfount, could I get a code review?

jniles:
@DedrickEnc This is ready for a third round of testing.

I've implemented the proposal in #797 for the stock expiration report, and linked it to the stock dashboard.  There were some unexpected hacks involved, but it works.  We can clean these up as we experiment with the deep linking pattern.

Stock Reversing needs to be implemented -- it's not entirely clear to me how it used to work, and the journal interface is difficult to work with.  I suggest that fixing it be a separate pull request.

jniles:
@DedrickEnc, this is ready for a fourth round of testing.

Some of the integration bugs reported last time are because the purchase table does not have a foreign key with currency_id, allowing both donations and integrations to be inserted with incorrect currency ids.

PR #767, opened 2015-09-24
Fixing texts in purchase order validation and authorization

mbayopanda:
In this PR, we have fixed the text in button which exceed the allocated zone

### DIFF
--- a/server/test/guid.spec.js
+++ b/server/test/guid.spec.js
@@ -3,7 +3,7 @@
 var expect = require('chai').expect,
     guid      = require('../lib/guid');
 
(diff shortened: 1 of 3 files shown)


server/test/guid.spec.js
```diff
@@ -3,7 +3,7 @@
 var expect = require('chai').expect,
     guid      = require('../lib/guid');
 
-describe('guild', function () {
+describe('guid', function () {
```
jniles: Thanks :)

jniles:
LGTM

PR #769, opened 2015-09-24
Fixing stock loss error

mbayopanda:
In this PR we have fixed an error which was occurring when we make a stock loss operation


jniles:
LGTM

PR #770, opened 2015-09-24
Required label for address area in supplier management

mbayopanda:
In this PR we have fixed label for an adress area in supplier management because when we try to edit a supplier we can't see that the address is required, and we cannot edit anything (the button is disabled).

### DIFF
--- a/client/src/partials/creditor/creditor.html
+++ b/client/src/partials/creditor/creditor.html
@@ -62,7 +62,7 @@
               </div>
 
               <div class="form-group">
-                <label for="ad1" class="control-label required">{{ "SUPPLIER.ADDR_1" | translate }}</label>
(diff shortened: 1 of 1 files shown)


client/src/partials/creditor/creditor.html
```diff
@@ -62,7 +62,7 @@
               </div>
 
               <div class="form-group">
-                <label for="ad1">{{ "SUPPLIER.ADDR_1" | translate }}</label>
+                <label for="ad1" class="control-label required">{{ "SUPPLIER.ADDR_1" | translate }}</label>
```
jniles: You don't need the `control-label` class here.  The `control-label` tag is used with `form-horizontal`, or when using form validation (which this module does not use).  If you would like, I can show you how to implement form validation tomorrow.

jniles:
LGTM.

Issue #329, opened 2015-01-14
Remove journal.js dependencies on sanitize.js


jniles:
Since we now encapsulate secure sanitation tools in db.exec, the next step is to remove journal.js's dependency on sanitize.js.  db.exec's formatting is much more readable and clear.  This may be a good first step to breaking journal.js into many separate components, as well as catching any bugs we may have during posting.

PR #771, opened 2015-09-24
Debtor grouppdf  report

DedrickEnc:
this PR add a funtionnality which help the enterprise to build the financial situation of a debitor group.
I have one or two thing to add just about CSS.


client/src/js/app.js
```diff
@@ -719,6 +719,10 @@ function bhimaconfig($routeProvider) {
     controller : 'configureBilan',
     templateUrl : 'partials/reports_proposed/bilan/bilan.html'
   })
+  .when('/reports/debitor_group_report/', {
+    controller : 'configureDebitorGroupReport as configureDebitorGroupReportCtrl',
```
jniles: It might be nicer to use a shorter name that is easier to read in the 'as' statement.

client/src/partials/reports_proposed/debitor_group_report/debitor_group_report.js
```diff
@@ -0,0 +1,76 @@
...
+    configurationObject.dg = vm.debitorGroup.get(vm.debitor_group_uuid);
+    configurationObject.unsoldOnly = vm.unsold_only;
+
+    // Update state
+    vm.building = true;
+
+    $http.post(path, configurationObject)
+    .success(function (result) {
+      // Expose generated document path to template
+      console.log('voici le result', result);
```
jniles: Could you remove these `console.log()`s?

client/src/partials/reports_proposed/debitor_group_report/debitor_group_report.js
```diff
@@ -0,0 +1,76 @@
...
+    configurationObject.project = configuration.project;
+    configurationObject.involveJournal = vm.involve_journal;
+    configurationObject.dg = vm.debitorGroup.get(vm.debitor_group_uuid);
+    configurationObject.unsoldOnly = vm.unsold_only;
+
+    // Update state
+    vm.building = true;
+
+    $http.post(path, configurationObject)
+    .success(function (result) {
```
jniles: The $http.success() object is [officially deprecated](https://docs.angularjs.org/api/ng/service/$http).  AngularJS will be removing it soon.  You should use `then/catch` promise methods here instead.

server/controllers/reports_proposed/data/debitor_group_report.js
```diff
@@ -0,0 +1,77 @@
+// reports_proposed/data/balance_sheet.js
+// Collects and aggregates data for the enterprise balance sheet
+var q       = require('q');
+var db      = require('../../../lib/db');
+var numeral = require('numeral');
+var sanitize = require('../../../lib/sanitize');
```
jniles: We are trying to remove sanitize.js.  (See #329).  Could you please remove all code that references it?

server/controllers/reports_proposed/data/debitor_group_report.js
```diff
@@ -0,0 +1,77 @@
...
+      item.difference = item.cost - item.credit;
+      item.cost = numeral(item.cost).format(formatDollar);      
+      item.credit = numeral(item.credit).format(formatDollar);
+      item.difference = numeral(item.difference).format(formatDollar);
+    });
+    context.somBilled = numeral(total.billed).format(formatDollar);
+    context.somPayed = numeral(total.payed).format(formatDollar);
+    context.somDifference = total.billed - total.payed;
+    context.somDifference = numeral(context.somDifference).format(formatDollar);
+    console.log('resultat', context.data);
```
jniles: Please remove the console.log()s.

server/controllers/reports_proposed/data/debitor_group_report.js
```diff
@@ -0,0 +1,77 @@
...
+    deferred.resolve(context);
+  })
+  .catch(deferred.reject)
```
jniles: Can't you just return db.exec()?  It is a promise.  Why do you need to create a new deferred promise?

server/controllers/reports_proposed/reports.js
```diff
@@ -74,6 +79,9 @@ exports.build = function (req, res, next) {
 
   var handler = documentHandler[target]; //handler will contain a object with two property, template for structure and context for data
   var options = req.body;
+  options.path = path.join(__dirname);
+
+  console.log('on est la ', options);
```
jniles: Remove this console.log().

server/controllers/reports_proposed/data/debitor_group_report.js
```diff
@@ -0,0 +1,77 @@
...
+var debitorGroupReportDate = new Date();
+
+exports.compile = function (options) {
+  'use strict';
+
+  var i18nDebitorGroupReport = options.language == 'fr' ? require('../lang/fr.json').DEBITOR_GROUP_REPORT : require('../lang/en.json').DEBITOR_GROUP_REPORT;
+  var deferred = q.defer(), context = {}, params = options.involveJournal === true ? [options.dg.account_id, options.dg.account_id] : [options.dg.account_id];
+
+  var sql2 =
+    'UNION ALL SELECT `s`.`uuid` as `saleUUID`, CONCAT(`pr`.`abbr`, `s`.`reference`) AS `saleNumber`, `s`.`cost`, DATE_FORMAT(`s`.`invoice_date`, \'%d-%m-%Y\') AS `invoice_date`, `dg`.`name`, `ac`.`account_txt`, SUM(`pj`.`debit_equiv`) AS `debit`, SUM(`pj`.`credit_equiv`) AS `credit`, `pj`.`trans_id`, ' +
```
jniles: In Node v4.0.0 (which we are running) the "\`" mark has a special meaning -- it is a [new type of string](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/template_strings) called a template string.  To make this code clear and maintainable, you should remove the "\`" marks from within the string.  MySQL does not require them, and it will improve code clarity.

[Here is another article on template strings](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Text_formatting)

server/models/development/update/synt.sql
```diff
@@ -271,4 +271,13 @@ ALTER TABLE  `cost_center`
...
\ No newline at end of file
+DROP FOREIGN KEY  `service_ibfk_1` ;
+
+--
+-- Dedrick Kitamuka
+-- 22/09/2015
+-- Adding a new unit for resultat comptable
+--
+
+INSERT INTO unit (`id`, `name`, `key`, `description`, `parent`, `has_children`, `url`, `path`)
```
jniles: The pull request (#670) from @lomamech removed the has_children column from the unit table.  I haven't tested this branch, but this code might crash because of it.  Could you remove the has_children column from this query?

server/controllers/reports_proposed/data/debitor_group_report.js
DedrickEnc: I am doing additional processing after fetching data, returning db.exec will return fetched data directly, that why I need a deferred promise.

server/controllers/reports_proposed/data/debitor_group_report.js
DedrickEnc: Got it, however mysql requires `` for key words, I removing that since there is no key word in the request string.

jniles: We can get around that by always name spacing (which is clearer anyway).  For example: `SELECT table.column` rather than `SELECT column`.

server/controllers/reports_proposed/data/debitor_group_report.js
jniles: I think you can use [promise chaining](http://www.html5rocks.com/en/tutorials/es6/promises/#toc-chaining).  If you return db.exec, and then return from inside the .then() statement, it should give you the results you want.

[Here's the q documentation.](https://github.com/kriskowal/q#chaining)

client/src/partials/reports_proposed/debitor_group_report/debitor_group_report.js
```diff
@@ -0,0 +1,74 @@
...
+        tables : {
+          'debitor_group' : {
+            columns : ['uuid', 'name', 'account_id', 'is_convention']
+          }
+        },
+        where : ['debitor_group.is_convention=1']
+      }
+  };
+
+  vm.generate_doc = $translate.instant('DEBITOR_GROUP_REPORT.GENERATE_DOC');
```
jniles: Instead of using $translate.instant() here (which will not update if the translation changes), why not put this in the view?  It would go on [this line](https://github.com/IMA-WorldHealth/bhima/pull/771/files#diff-27d9fb20a3fcd83f5250cfbf4ae2e384R59), with the same logic.

client/src/partials/reports_proposed/debitor_group_report/debitor_group_report.js
```diff
@@ -0,0 +1,74 @@
...
+    configurationObject.involveJournal = vm.involve_journal;
+    configurationObject.dg = vm.debitorGroup.get(vm.debitor_group_uuid);
+    configurationObject.unsoldOnly = vm.unsold_only;
+
+    // Update state
+    vm.building = true;
+
+    $http.post(path, configurationObject)
+    .then(function (result) {
+      console.log(result);
```
jniles: Please remove this console.log().

server/controllers/reports_proposed/data/debitor_group_report.js
```diff
@@ -0,0 +1,75 @@
...
+var formatDollar = '$0,0.00';
+var debitorGroupReportDate = new Date();
+
+exports.compile = function (options) {
+  'use strict';
+
+  var i18nDebitorGroupReport = options.language == 'fr' ? require('../lang/fr.json').DEBITOR_GROUP_REPORT : require('../lang/en.json').DEBITOR_GROUP_REPORT;
+  var deferred = q.defer(), context = {}, params = options.involveJournal === true ? [options.dg.account_id, options.dg.account_id] : [options.dg.account_id];
+
+  var sql2 =
```
jniles: Could you use a more meaningful name than sql2? Also, why does sql2 come before sql in the definitions?

server/controllers/reports_proposed/data/debitor_group_report.js
```diff
@@ -0,0 +1,75 @@
...
+    'UNION ALL SELECT s.uuid as saleUUID, CONCAT(pr.abbr, s.reference) AS saleNumber, s.cost, DATE_FORMAT(s.invoice_date, \'%d-%m-%Y\') AS invoice_date, dg.name, ac.account_txt, SUM(pj.debit_equiv) AS debit, SUM(pj.credit_equiv) AS credit, pj.trans_id, ' +
+    'CONCAT(pt.first_name, \' \', pt.last_name) AS patientName FROM sale AS s JOIN posting_journal AS pj ON pj.inv_po_id = s.uuid JOIN account AS ac ON pj.account_id = ac.id JOIN debitor_group AS dg ' +
+    'ON dg.account_id = pj.account_id JOIN patient AS pt ON pt.debitor_uuid = s.debitor_uuid JOIN project AS pr ON pr.id = s.project_id WHERE pj.account_id =? GROUP BY s.uuid';
+  var addSql = options.involveJournal === true ? sql2 : '';
```
jniles: Should we really offer the option to see the posting journal in this report?  For an official report (which a PDF report is), it is incorrect accounting to use the posting journal.

server/controllers/reports_proposed/data/debitor_group_report.js
```diff
@@ -0,0 +1,75 @@
...
+  var sql = 
+    'SELECT s.uuid as saleUUID, CONCAT(pr.abbr, s.reference) AS saleNumber, s.cost, DATE_FORMAT(s.invoice_date, \'%d-%m-%Y\') AS invoice_date, dg.name, ac.account_txt, SUM(gl.debit_equiv) AS debit, SUM(gl.credit_equiv) AS credit, gl.trans_id, ' +
+    'CONCAT(pt.first_name, \' \', pt.last_name) AS patientName FROM sale AS s JOIN general_ledger AS gl ON gl.inv_po_id = s.uuid JOIN account AS ac ON gl.account_id = ac.id JOIN debitor_group AS dg ' +
+    'ON dg.account_id = gl.account_id JOIN patient AS pt ON pt.debitor_uuid = s.debitor_uuid JOIN project AS pr ON pr.id = s.project_id WHERE gl.account_id =? GROUP BY s.uuid ' + addSql;
+  var filterResults = function (results){
```
jniles: It is better to use named function definitions (such as `function filterResults() { /* ... code .. */ }`) rather than anonymous functions (such as `var x = function () { /* ... code ... */ };`).  Not only is the first option more flexible, since it is hoisted to the top of the scope, any errors will be easier to debug (the error will print the function name it was thrown in).

server/controllers/reports_proposed/data/debitor_group_report.js
```diff
@@ -0,0 +1,75 @@
...
+  context.amountBilled = i18nDebitorGroupReport.AMOUNT_BILLED;
+  context.amountPayed = i18nDebitorGroupReport.AMOUNT_PAYED;
+  context.difference = i18nDebitorGroupReport.DIFFERENCE;
+  context.transNumber = i18nDebitorGroupReport.TRANS_NUMBER;
+  context.date = i18nDebitorGroupReport.DATE;
+  context.totalBilled = i18nDebitorGroupReport.TOTAL_BILLED;
+  context.totalPayed = i18nDebitorGroupReport.TOTAL_PAYED;
+  context.totalDifference = i18nDebitorGroupReport.TOTAL_DIFFERENCE;
+  context.tel = i18nDebitorGroupReport.TEL;
+  context.box = i18nDebitorGroupReport.BOX
```
jniles: Is it possible to reduce the number of lines here?  Something like `var context = i18nDebtorGroupReport;`, rather than transferring over each property individually?

server/controllers/reports_proposed/data/debitor_group_report.js
```diff
@@ -0,0 +1,75 @@
...
+      item.difference = item.cost - item.credit;
+      item.cost = numeral(item.cost).format(formatDollar);      
+      item.credit = numeral(item.credit).format(formatDollar);
+      item.difference = numeral(item.difference).format(formatDollar);
+    });
+    context.somBilled = numeral(total.billed).format(formatDollar);
+    context.somPayed = numeral(total.payed).format(formatDollar);
+    context.somDifference = total.billed - total.payed;
+    context.somDifference = numeral(context.somDifference).format(formatDollar);
+    deferred.resolve(context);
```
jniles: I'm pretty sure you can just `return db.exec().then(function (results) { /* ... /* return context; });`.

server/controllers/reports_proposed/reports.js
```diff
@@ -74,6 +80,7 @@ exports.build = function (req, res, next) {
 
   var handler = documentHandler[target]; //handler will contain a object with two property, template for structure and context for data
   var options = req.body;
+  options.stylePath = path.join(__dirname);
```
jniles: This can be `options.stylePath = __dirname`;  Although, I would just name it `path` or something, since it doesn't have anything to do with styling (even if you use it later for styling).

server/controllers/reports_proposed/templates/style/reports.css
```diff
@@ -39,11 +39,15 @@ tr.outline {
...
 
-.center > * {
+.header_left {  
+  float: left;
+  width: 40%
+}
+
+.header_center {  
+  float: left;
+  width: 40%;
```
jniles: header_center === header_left, except for `text-align:center`.  It would probably be better to create an addon class (`.center-text`, for example) that you can just combine in the HTML, rather than create a whole new class.

server/controllers/reports_proposed/templates/style/reports.css
```diff
@@ -39,11 +39,15 @@ tr.outline {
   border-bottom : 1px solid #9f9f9f;
 }
 
-.center > * {
+.header_left {
```
jniles: Traditionally, it has been considered bad practice to use underscores in CSS class names.  In fact, for old browsers, it used to break CSS ([source](https://developer.mozilla.org/en-US/docs/Web/Guide/CSS/Underscores_in_class_and_ID_Names)).

[Here is a poll on CSS tricks on using hyphens (`-`), underscores (`_`) or camelCase for class names](https://css-tricks.com/new-poll-hyphens-or-dashes/).  People seem to favor either dashes (like in bootstrap) or camelCase.

server/controllers/reports_proposed/data/debitor_group_report.js
```diff
@@ -0,0 +1,75 @@
+// reports_proposed/data/balance_sheet.js
+// Collects and aggregates data for the enterprise balance sheet
```
jniles: These comments are misleading and wrong.  Either change them or remove them.

client/src/partials/reports_proposed/debitor_group_report/debitor_group_report.html
```diff
@@ -0,0 +1,66 @@
...
+            <input type="checkbox" ng-value=1 ng-model="debtorGroupReportCtrl.unsold_only">
+          </div>
+        </div>
+
+        <div class="form-group">
+          <label>
+            {{'DEBITOR_GROUP_REPORT.LANGUAGE' | translate}}
+          </label>
+
+          <div>
```
jniles: Should this have a class="form-group" on it?

client/src/partials/reports_proposed/debitor_group_report/debitor_group_report.js
```diff
@@ -0,0 +1,74 @@
...
+      vm.generatedDocumentPath = result.data;
+    })
+    .catch(function (err) {
+      vm.building = false;
```
jniles: Should `vm.building` be set to false after the $http request completes in both the `.then()` and the `.catch()` methods?  If so, why not put it in a `.finally()` method?

server/models/development/update/synt.sql
```diff
@@ -271,4 +271,13 @@ ALTER TABLE  `cost_center`
...
 
 ALTER TABLE  `service` 
-DROP FOREIGN KEY  `service_ibfk_1` ;
\ No newline at end of file
+DROP FOREIGN KEY  `service_ibfk_1` ;
+
+--
+-- Dedrick Kitamuka
+-- 22/09/2015
+-- Adding a new unit for resultat comptable
```
jniles: Is this comment correct?

client/src/partials/reports_proposed/debitor_group_report/debitor_group_report.js
DedrickEnc: $translate service works fine and changes when the translation changes.
I avoid to use it on the view because, it will make the html page unreadable (putting conditional structure on a html page is not catholic).

jniles: I'm almost positive that it only updates translation because the controller is re-run.  Although, [the documentation](https://angular-translate.github.io/docs/#/api/pascalprecht.translate.$translate) does not make it clear if $translate.instant() is scheduled to update on language change.  However, if you read the [source code](https://github.com/angular-translate/angular-translate/blob/920a9fe/src/service/translate.js#L1935), it looks like the instant method does not schedule listeners to be run multiple times.

You already have logic in the view...at this very spot.  But if you do not want to put translation logic in the view, just do
```js
vm.message = vm.loading ? 'DEBITOR_GROUP_REPORT.GENERATE_DOC' : 'DEBITOR_GROUP_REPORT.LOADING';
```

and update accordingly.

server/controllers/reports_proposed/data/debitor_group_report.js
DedrickEnc: I disagree, It is not good to avoid giving possibility to involve posting_journal record when necessary.
That is an option and not an obligation.

jniles: I would talk to Mr. Wayne about it.  We've already seen at our test site that people will not post if we give them reports on the posting_journal.

server/controllers/reports_proposed/data/debitor_group_report.js
DedrickEnc: Got it

server/controllers/reports_proposed/data/debitor_group_report.js
DedrickEnc: That instruction will overwrite the context object. I will make context object deep by using : 
```
context.dataStructure = i18nDebtorGroupReport;
```

jniles: Sounds good.

client/src/partials/reports_proposed/debitor_group_report/debitor_group_report.html
DedrickEnc: Necessary to have good space between controls.

jniles:
These changes are now really good.  There are two things that need to be done before this PR will be merged in:
 1. You should `git rebase -i` and remove the merges and commits that are 'fixing conflicts'.
 2. I really strongly feel that we cannot allow the posting journal in reports.  Our test site has already demonstrated that they will NOT post if they can see data from the posting journal.  Additionally, reports furnish information with which the administration can make decisions.  Because the posting journal is not guaranteed to be correct, we should not even provide it as an option -- it will likely lead to incorrect decisions.

DedrickEnc:
I tried to do git rebase, and I expend about 30 min to fix conflict and some conflicts was unclear to resolve (Jonathan Cameron add this problem when sending budget PR some month ago), I did the merge because it was the easiest way to do that (even if there is one or two unnecessary commits, I saved time and keep the code safe.).
People at Tshikaji don't post because they can not post not because they don't want to post.
The general ledger is the only source of information when generating synthetic report of accounting such as bilan, balance, TAFIR and so one.

Here we are talking about a financial report which can be spontaneous, even if it can be official (In this case, the user should choose general ledger only).
We have the similar situation when using auxiliary cash, we are fetching data from the posting journal and general ledger  which is fine for me instead of removing this option we should let the user know that there is a risk to have bad data and also this module deal with our very stable module which are sale and convention payment.

One more thing, we should think about risk of being unable to post data to the general ledger after the trip and users must be able to present representative report to their client if the are confident about validity of their data.

jniles:
Auxiliary cash is not a report.  It's looking for unbalanced invoices to pay.  If it comes up with an incorrect amount, the patient with his/her sale receipt will say "this is not right" and we can find a different way to charge him.  There is no way to double-check a report like this.

If this is an unofficial report, I wouldn't have done it in PDF.  I was under the impression this was something official we would send to a convention.

Okay, not a problem with the git rebase.  We should all probably learn a bit more about git so that we can use it effectively.

I guess this has my OK to go into the development branch.

PR #773, opened 2015-09-28
Fixing fiscal year closing

mbayopanda:
In this PR we have fixed : 
  - An error which occur when we try to close a fiscal year
  - Writing in the journal relative data of the fiscal year closed

### DIFF
--- a/server/controllers/journal/fiscal.js
+++ b/server/controllers/journal/fiscal.js
@@ -3,7 +3,8 @@ var q         = require('q'),
     uuid      = require('../../lib/guid'),
     sanitize  = require('../../lib/sanitize'),
     validate  = require('../../lib/validate')(),
-    db        = require('../../lib/db');
+    db        = require('../../lib/db'),
+    util      = require('../../lib/util');
 
 exports.close = close;
 exports.create = create;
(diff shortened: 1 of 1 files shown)


server/controllers/journal/fiscal.js
```diff
@@ -50,11 +51,11 @@ function close(id, user_id, data, cb) {
   function init() {
     cfg.user_id = user_id;
     cfg.project_id = 1; // HBB by default
-    transactionDate = cfg.isForClosing ? forcingDate : new Date();
-    return [
+    transactionDate = cfg.isForClosing ? util.toMysqlDate(forcingDate) : util.toMysqlDate();
+    return q.when([
```
jniles: Do you really need q.when()?  [According to the Q documentation](https://github.com/kriskowal/q#combination), Q will create a new promise (using Q.all) internally, so you do not need to create one here.

mbayopanda: I needed a promise because after the init() function there is a ".spread(...) ". The error was that the init() function returns a simple array instead of a promise or if it was in a promise chain

jniles: Aha.  Didn't realize init() was the first in the chain.  Thanks for clearing that up.  LGTM.

Issue #719, opened 2015-09-09
Adopt 'Controller As' Syntax and Deprecate $scope


jniles:
I"m putting this one up for discussion and visibility.  In my opinion, this is the way forward.

Angular gives us the 'Controller As' syntax to bind a controller directly to the DOM.  It looks like this:
```js
angular.module('ExampleApp', []);

function routeConfig($routeProvider) {
  $routeProvider.when('/', {
    controller : 'ExampleController as ExampleCtrl',
    templateUrl : 'example.html'
  });
}

angular.config(['$routeProvider', routeConfig]);
```

It is currently used in [login.js](https://github.com/IMA-WorldHealth/bhima/blob/development/client/src/partials/auth/login.js) and [login.html](https://github.com/IMA-WorldHealth/bhima/blob/development/client/src/partials/auth/login.html), if you would like to see it in a live application.  You can either explicitly use ` { controller : 'Controller as Alias' }` or use `{ controller : 'Controller', controllerAs : 'Alias' }`.

What this does is bind the controller to the DOM, so that you can use the `this` keyword in place of $scope.  This commonly looks like this:

```js
angular.module('ExampleApp')
.controller('ExampleController', ExampleController);

function ExampleController() {
  var vm = this;  // vm stand for view-model
  vm.hello = 'world!'
}
```
... and in the DOM...
```html
<!-- example.html -->
<section>{{ ExampleCtrl.hello }}</section>
<!-- this will render to <section>world!</section> -->
```

Why abandon $scope?  To me, $scope is an unclear dependency on Angular.  It is verbose, and difficult to type (that $).  If our purpose is to write clean code, $scope is not helping us.  Additionally, we have all used `$scope.session` or `$scope.params` or `$scope.data` or `$scope.models` to try and get around the issue of not having a dot (object.property) in the view.  By binding to the controller, we always have the top-level controller alias in the reference `{{ alias.property }}`.

Nested $scopes are also clearer to debug in the HTML.  For example,
```html
<section>
   <h3> {{ parent.property }} </h3>
   <div ng-controller="ChildController as child">
     {{ child.property }}
   </div>
</section>
```
The code is self-describing.  We have these relationships in our index.html file, and some of the files that load other templates via `ng-include`.

[...]

Issue #679, opened 2015-09-02
Multiple payroll UI is slow


jniles:
The multiple payroll page could use some optimizations.  When you load up the page, it takes a very long time to template in the employee data if you have many employees.  (Hint: we currently have **148** employees).  For each employee, we template in an entire panel.

Perhaps the entire UI design needs to be re-evaluated -- but it would be good (at least, temporarily) to find a way to make sure the client does not block as long as it currently does.

PR #776, opened 2015-09-30
Budget importation via csv

mbayopanda:
In this PR : 
- We use : 
  - npm modules : "fast-csv", "connect-multiparty" 
  - angular module : "ng-file-upload"

The main task in this PR is to import budget from csv file, and make the imported budget the new budget of the enterprise.

We have to : 
- make "npm install" for installing 'fast-csv' and 'connect-multiparty'
- make "bower install" for installing 'ng-file-upload'

In this PR, i fixed some codes in budget analysis : 
- The label of the csv file generated
- The ui of budget analysis for harmony with "budget management"


client/src/index.html
```diff
@@ -114,6 +114,10 @@
 <!--locale-->
 <script src="vendor/angular-dynamic-locale/dist/tmhDynamicLocale.js"></script>
 
+<!-- ng-file-upload -->
+<script src="vendor/ng-file-upload-shim/ng-file-upload-shim.min.js"></script> <!-- for no html5 browsers support -->
```
jniles: I'm pretty sure we don't need this file -- although that decision can be made when we discuss installation.  At the moment, we only "officially" support the latest chrome, which supports HTML5.

client/src/js/app.js
```diff
@@ -19,6 +19,10 @@ function bhimaconfig($routeProvider) {
     controller: 'budget.analysis',
     templateUrl: 'partials/budget/analysis/analysis_budget.html'
   })
+  .when('/budgeting/new', {
+    controller: 'budget.new',
```
jniles: Could you use 'controller as' syntax?

client/src/partials/budget/analysis/analysis_budget.js
```diff
@@ -360,7 +360,9 @@ angular.module('bhima.controllers')
       });
 
       // Exportation
-      exportFile.csv(fileData, 'Budget Analysis Report');
+      var labelPeriod     = session.selectedPeriod || '';
+      var labelFiscalYear = session.selectedFiscalYear.start_year || '';
```
jniles: Should you allow these to be undefined or false?  It is better to default to `''` than throw an error?

client/src/partials/budget/new/new_budget.js
```diff
@@ -0,0 +1,368 @@
...
+  '$http',
+  '$translate',
+  'validate',
+  'precision',
+  'messenger',
+  'appstate',
+  'SessionService',
+  'connect',
+  'Upload',
+  function($q, $scope, $window, $http, $translate, validate, precision, messenger, appstate, SessionService, connect, Upload) {
```
jniles: In #719 we decided to use 'controller as' syntax and `controller.$inject` to manage AngularJS dependency injection.  Could you please rewrite this code to reflect that?

client/src/partials/budget/new/new_budget.js
```diff
@@ -0,0 +1,368 @@
...
+
+    function createBudget() {
+      var newBudgets = [];
+      $scope.periods.data.forEach(function (per, index) {
+      newBudgets.push({'account_id' : session.account.id,
+       'period_id' : per.id,
+       'budget' : 0.0});
+      });
+
+      connect.post('budget', newBudgets, ['id'])
```
jniles: `connect.post()` only takes two parameters.  Why did you use three?

client/src/partials/budget/new/new_budget.js
```diff
@@ -0,0 +1,368 @@
...
+      console.log(err);
+      });
+    }
+
+    function updateBudget() {
+      // Save the budget data for all the periods
+      var dbPromises = [];
+      $scope.budgets.data.forEach(function (bud) {
+      dbPromises.push( 
+      connect.put('budget',
```
jniles: Here, you are generating a loop of HTTP requests.  This is exactly the problems we reported on #624, #679 and others.

A better model is to create a server-side route that receives one HTTP request and send all the data at once to that endpoint.  It can then validate the data and insert it into the database.

client/src/partials/budget/analysis/analysis_budget.js
mbayopanda: It's less probable to have session.selectedPeriod or session.selectedFiscalYear to undefined but if this case occur, it'll be better to have ' ' for both, and the name of the file will be like "Budget 2015-09-30-14-57-00"

client/src/upload/description.md
```diff
@@ -0,0 +1,4 @@
+Upload folder
```
jniles: You should add this folder (but not this file) to `.gitignore`

client/src/partials/budget/new/new_budget.js
```diff
@@ -0,0 +1,361 @@
...
+
+    // BUDGET: 
+    //   `id` int not null auto_increment,
+    //   `account_id` int unsigned not null default '0',
+    //   `period_id` mediumint unsigned not null,
+    //   `budget` decimal(10,4) unsigned,
+
+    function createBudget() {
+      var newBudgets = [];
+      $scope.periods.data.forEach(function (per, index) {
```
jniles: Could you indent this code properly?  Each new block should be two spaces.

The current formatting makes it difficult to read.

client/src/partials/budget/new/new_budget.js
```diff
@@ -0,0 +1,361 @@
...
+              bud.budget = totalFree / numFree;
+            }
+          });
+          session.validTotal = true;
+        }
+      }
+
+      // Double-check the totals
+      var total = 0.0;
+      $scope.budgets.data.forEach(function (bud) {
```
jniles: It would be much nicer to use an array.reduce()!

server/controllers/budget.js
```diff
@@ -0,0 +1,123 @@
+/* Budget controller */
+'use strict';
+
+var csv  = require('fast-csv'),
+		fs   = require('fs'),
```
jniles: Please fix this indentation.

server/controllers/budget.js
```diff
@@ -0,0 +1,123 @@
...
+		db   = require('../lib/db'),
+		q    = require('q');
+
+var uploadDir = 'client/upload/';
```
jniles: This should be in a config file.  Probably `server/config/environment`.

client/src/partials/budget/new/new_budget.html
```diff
@@ -0,0 +1,197 @@
...
+        <li class="active">{{ "BUDGET.NEW.TITLE" | translate }}</li>
+      </ol>
+  </div>
+</nav>
+
+<div class="report">
+  <div ng-switch on="session.mode">
+    <div ng-switch-when="configure">
+      <div class="col-sm-offset-3 col-xs-6">
+        <h3 class="center"><b>{{ "BUDGET.NEW.TITLE" | translate }}</b></h3>
```
jniles: I'm pretty sure the `<b>

client/src/partials/budget/new/new_budget.html
```diff
@@ -0,0 +1,197 @@
...
+        <p>{{ "BUDGET.NEW.DESC" | translate }}</p>
+
+        <form class="form budget-analysis-form" novalidate>
```
jniles: There is no class `form`, unless you have defined it elsewhere.  It's a common mistake I make.

client/src/partials/budget/analysis/analysis_budget.js
jniles: Okay, sound good.

client/src/partials/budget/new/new_budget.html
```diff
@@ -0,0 +1,199 @@
...
+          </div>
+        </div>
+      </div>
+
+      <div class="row" ng-if="session.found">
+        <div class="panel panel-default">
+          <div class="panel-body">
+            <div class="col-xs-8">
+              <form name="budget">
+                <table class="table table-condensed form-bhima-border" summary="{{ 'EDIT.BUDGET.EDIT' | translate }}">
```
jniles: Table `summary` seems like a useful attribute, but according to the [Mozilla Developer Network](https://developer.mozilla.org/en/docs/Web/HTML/Element/table), it is deprecated.  Here's what the Mozilla developers recommend:
> Usage Note: Do not use this attribute, as it has been deprecated. Instead, use one of these way of describing a table:
In prose, surrounding the table (this is the less semantic-conveying way of doing it).
In the table's `<caption>` element.
In a `<details>` element, inside the table's `<caption>` element.
Include the `<table>` element in a `<figure>` element and add the description in prose next to it.
Include the `<table>` element in a `<figure>` element and add the description in prose inside a `<figcaption>` element.
Adjust the table so that such description is no more needed, by using `<th>` and `<thead>` elements for example.
> -- Mozilla Developer Network

client/src/partials/budget/new/new_budget.html
```diff
@@ -0,0 +1,199 @@
...
+                      <td><input type="number" 
+                           ng-model="budget.budget" 
+                           ng-change="recompute()" 
+                           ng-focus="startEditing(budget)"
+                           ng-blur="endEditing(budget)"
+                           min="0.0" 
+                           ng-disabled="session.totalBudget <= 0.0 || (session.autoAdjust && (budget.freeze && !budget.editing))"
+                           >
+                      </td>
+                      <td style="text-align: center">
```
jniles: Use `class='text-center'` to make the code more uniform.

client/src/partials/budget/new/new_budget.html
```diff
@@ -0,0 +1,199 @@
...
+                        <input type="checkbox" 
+                         style="margin: 0"
+                         ng-disabled="session.totalBudget <= 0.0"
+                         ng-change="recompute()"
+                         ng-model="session.autoAdjust"></b>
+                      </td>
+                      <td style="vertical-align: middle">
+                        <span class="pull-right">{{ "BUDGET.EDIT.TOTAL" | translate }}</span>
+                      </td>
+                      <td style="vertical-align: middle">
```
jniles: You have used `style="vertical-align:middle"` three times in this table on `<td>`s.  Could this be applied to the `<tr>` or make into a class that does this to avoid repetition?

client/src/partials/budget/new/new_budget.html
```diff
@@ -0,0 +1,199 @@
...
+    </div> <!-- end div session.mode = 'edit' -->
+  </div> <!-- end div session.mode -->
+</div> <!-- end div report -->
+
+<footer>
+  <div class="pull-right" ng-if="session.mode === 'edit'">
+    <button class="btn btn-default" 
+      style="margin-right: 20px"
+      ng-click="restartSearch()">
+      <span class="glyphicon glyphicon-refresh"></span> {{ "BUDGET.EDIT.RESTART_SEARCH" | translate }}</button>
```
jniles: Put `</button>` on a new line for readability.

client/src/partials/budget/new/new_budget.js
```diff
@@ -0,0 +1,357 @@
+angular.module('bhima.controllers')
+.controller('NewBudgetController', [
```
jniles: You should use the `$inject` property for readability and testability.  @sfount suggested it in #719.

Here's how to touch up the code to abide by this convention:
```js
.controller('NewBudgetController', NewBudgetController); // we will define this further down as a named function

NewBudgetController.$inject = [
  '$q', '$scope', '$http',  /* all other dependencies go here ... */
];

/**
*  New Budget Controller
*  This controller is responsible for creating a new budget.  Keep in mind to write useful comments in this space!
*  You do not need to discuss the parameters (since they are injected and can be lookup up as services), but make
*  sure you describe what the controller does.
*/
function NewBudgetController($q, $scope, $http, /* other dependencies ... */) {
  /* Awesome code */
}
```

mbayopanda:
In the "Use of in controllers", git remove "$inject" from message. The initial message was "Use of $inject in controllers" :-)

jniles:
You can always run `git commit --amend` to fix the last commit message!

server/config/routes.js
```diff
@@ -37,9 +37,13 @@ var accounts        = require('../controllers/accounts');
...
     projects        = require('../controllers/projects'),
     users           = require('../controllers/users'),
-    analytics       = require('../controllers/analytics');
+    analytics       = require('../controllers/analytics'),
+    budget          = require('../controllers/budget');
 
-var patient = require('../controllers/patient');
+var patient         = require('../controllers/patient');
+
+// Middleware for a particular route
```
jniles: You might want to say "middleware for handling uploads".  The current documentation in not very specific.

client/src/partials/budget/new/new_budget.js
```diff
@@ -0,0 +1,365 @@
...
+
+    // Double-check the totals
+    var total = $scope.budgets.data.reduce(sumBudget, 0.0);
+
+    if (isNaN(session.totalBudget) || session.totalBudget === null ||
+      isNaN(total) || total === null) {
+      // Make sure we have real numbers
+      session.validTotal = false;
+    } else {
+      session.validTotal = precision.round(total, 6) === precision.round(session.totalBudget, 6);
```
jniles: In the previous two lines, session.validTotal can be a boolean or a number.  Remember that `0 == false` in javascript and that `!0 === true`.  You should keep your total (number) in one variable and the validity check (boolean) in another, or find a different way to do this.

client/src/partials/budget/new/new_budget.js
```diff
@@ -0,0 +1,365 @@
...
+      'period' : {
+        columns : ['fiscal_year_id', 'period_number', 'period_start', 'period_stop', 'locked' ]
+      }
+    },
+    join  : [ 'period.id=budget.period_id' ],
+    where : [ 'period.period_number<>0' ]
+    }
+  };
+
+  dependencies.periods = {
```
jniles: This is a duplication with the function loadPeriod.  In the loadPeriod function (line [137](https://github.com/IMA-WorldHealth/bhima/pull/776/files#diff-026166d2e4e2dd47428c20c9eccf2c53R137)) you set up `dependencies.period` (note: period !== periods, but they do very close to the same thing).  If you could combine the two requests (dependencies.period and dependencies.periods) in to the same request object, that would be much cleaner.

client/src/partials/budget/analysis/analysis_budget.js
```diff
@@ -9,412 +11,423 @@ angular.module('bhima.controllers')
...
+        },
+      },
+      orderby : ['fiscal_year.start_year', 'fiscal_year.start_month'],
+    }
+  };
+
+  init();
+
+  function init() {
+    enterprise_id = Number(SessionService.enterprise.id);
```
jniles: Is there a reason to have two variables: `enterprise_id` and `$scope.enterprise.id`?  Why not keep just `$scope.enterprise` and access the enterprise ID from there?

client/src/partials/budget/analysis/analysis_budget.js
```diff
@@ -9,412 +11,423 @@ angular.module('bhima.controllers')
...
+
+  function addBudgetData() {
+    // Insert the budget numbers into the account data
+    // TODO: The following procedural hacks can be simplified by better SQL queries...
+    // First compute the totals for any accounts with budgets
+    var totalBudget = 0.0;
+    var totalBalance = 0.0;
+    var totals = {};
+    $scope.budgets.data.forEach(function (bud) {
+      if (bud.account_id in totals) {
```
jniles: `for (expression in object) {` can behave strangely.  For example:
```js
var o = {};  assert('toString' in o === true); // this is true!
```
It is better to check if the property exists on the present object by using ` if (!totals[bud.account_id]) { /*.. */ }`.  For example,
```js
$scope.budgets.data.forEach(function (bud) {
  
  // initialize totals to zero
   if (!totals[bud.account_id]) { totals[bud.account_id] = 0 }
   
   // we can now add the totals without doing any other checks!
   totals[bud.account_id] += bud.budget;
   totalBudget += bud.budget;
});
```

client/src/partials/budget/new/new_budget.js
mbayopanda: session.validTotal variable is a boolean, and total variable is a number;
At the line 326, session.validTotal is still a boolean result of testing two numbers (total and session.totalBudget).

jniles: So it is.  Sorry.

client/src/partials/budget/new/new_budget.js
```diff
@@ -0,0 +1,357 @@
...
+
+  function updateBudget() {
+    // Save the budget data for all the periods
+    $http.post('/budget/update', { 
+      params : {
+        budgets   : $scope.budgets.data,
+        accountId : session.account.id
+      }
+    })
+    .success(function () {
```
jniles: The $http promise methods `.success()` and `.error()` are deprecated.  Do not use them.

client/src/partials/budget/new/new_budget.js
```diff
@@ -0,0 +1,357 @@
...
+  }
+
+  function sumBudget(a, b) {
+    return a + b.budget;
+  }
+
+  function loadFiscalYears(models) {
+    angular.extend($scope, models);
+    // Default to the last fiscal year
+    session.fiscal_year = $scope.fiscal_years.data[$scope.fiscal_years.data.length - 1];
```
jniles: This may become a problem later.  Is there anyway to default to the current fiscal year, _and then_ the last one, if we do not have a fiscal year?

server/controllers/budget.js
```diff
@@ -0,0 +1,125 @@
...
+		return db.exec(sql, [period_id]);
+	}
+}
+
+function updateBudget(req, res, next) {
+	var budgets = req.body.params.budgets;
+	var accountId = req.body.params.accountId;
+	var dbPromises = [];
+
+	budgets.forEach(function (bud) {
```
jniles: You can use the same `map()` style to transform this query as discussed [here](https://github.com/IMA-WorldHealth/bhima/pull/778/files#diff-2efedb2607b3ab9c464aab61e78fbab1R87).

server/controllers/budget.js
```diff
@@ -0,0 +1,125 @@
...
+
+	/*
+		* Objectif : create a new budget via params given
+		* Params : 
+		*	- csvArray : an array of the budget csv file
+		* - fiscal_year_id : the fiscal year id
+		* - period : the period
+	*/
+
+	if (csvArray.length) {
```
jniles: You can remove one level of indentation by doing this:
```js
if (!csvArray.length) { return; }
// the rest of your code
```

server/controllers/budget.js
```diff
@@ -0,0 +1,125 @@
...
+						});
+					}
+				});
+			});
+
+		} else if (period.period_number >= 1 && period.period_number <= 12) {
+			// Budget for a particular period in fiscal year
+			csvArray.forEach(function (item) {
+				if (item.Type !== 'title' && Number(item.Budget) !== 0 && !isNaN(Number(item.Budget))) {
+					var sql = 'INSERT INTO `budget` (account_id, period_id, budget) VALUES (?, ?, ?) ;';
```
jniles: You are re-defining the `sql` variable each turn of the loop.  Since it doesn't depend on the values within the loop, you should move it out of the loop.
```js
var sql = 'INSERT ... ';
csvArray.forEach(function (item) { /* code */ });
```

server/controllers/budget.js
```diff
@@ -0,0 +1,125 @@
...
+					var value = [item.AccountId, period.id, item.Budget];
+					clearPreviousBudgets(period.id)
```
jniles: How do you know these all complete?  Where is the error handling code?  This looks like it should have a `q.all()` to make sure everything actually succeeds before returning a success message.

server/controllers/budget.js
```diff
@@ -0,0 +1,125 @@
...
+
+	budgets.forEach(function (bud) {
+		var sql = 'UPDATE budget SET budget.account_id=?, budget.period_id=?, budget.budget=? WHERE budget.id=? ;';
+		var value = [accountId, bud.period_id, bud.budget, bud.id];
+		dbPromises.push(db.exec(sql, value));
+	});
+
+	q.all(dbPromises)
+	.then(function () {
+		res.status(200).send('update');
```
jniles: Where is the error handling?  If `q.all()` fails, **this HTTP request will never return to the client**.

server/controllers/budget.js
```diff
@@ -0,0 +1,125 @@
...
+		if (period.period_number === 0) {
+			// Budget for the fiscal year
+			var queryPeriods = 'SELECT period.id, period.period_number FROM period WHERE period.fiscal_year_id = ? ';
+			db.exec(queryPeriods, [fiscal_year_id])
+			.then(function (data) {
+				var periodIds = data;
+				var periodLength = periodIds.length - 1 > 0 ? periodIds.length - 1 : 1;
+				periodIds.forEach(function (item) {
+					if (item.period_number !== 0) {
+						csvArray.forEach(function (acc) {
```
jniles: This operation is identical to the operation defined on [line 87](https://github.com/IMA-WorldHealth/bhima/pull/776/files#diff-c657a4fbbd424edf858c073c94e8a238R87).  Maybe it is better to create a new function that inserts data into the budget table?

server/controllers/budget.js
```diff
@@ -0,0 +1,125 @@
...
+							// Hack for performance : Number(acc.Budget) !== 0 skip default budget=0
+							// If we choose to set a budget to zero in csv file, it's will be impossible to import it
+							if (acc.Type !== 'title' && Number(acc.Budget) !== 0 && !isNaN(Number(acc.Budget))) {
+								var sql = 'INSERT INTO `budget` (account_id, period_id, budget) VALUES (?, ?, ?) ;';
+								var value = [acc.AccountId, item.id, acc.Budget/periodLength];
+								clearPreviousBudgets(item.id)
+								.then(function () {
+									db.exec(sql, value);
```
jniles: Since this code is identical to the code below, the comments on error handling and using `q.all()` apply here as well.

server/controllers/budget.js
```diff
@@ -0,0 +1,125 @@
...
+
+	if (csvArray.length) {
+
+		if (period.period_number === 0) {
+			// Budget for the fiscal year
+			var queryPeriods = 'SELECT period.id, period.period_number FROM period WHERE period.fiscal_year_id = ? ';
+			db.exec(queryPeriods, [fiscal_year_id])
+			.then(function (data) {
+				var periodIds = data;
+				var periodLength = periodIds.length - 1 > 0 ? periodIds.length - 1 : 1;
```
jniles: There should be a comment describing this.  Why are you setting `periodLength` to anything else but `periodIds.length`?

server/controllers/budget.js
```diff
@@ -0,0 +1,125 @@
...
+		* - fiscal_year_id : the fiscal year id
+		* - period : the period
+	*/
+
+	if (csvArray.length) {
+
+		if (period.period_number === 0) {
+			// Budget for the fiscal year
+			var queryPeriods = 'SELECT period.id, period.period_number FROM period WHERE period.fiscal_year_id = ? ';
+			db.exec(queryPeriods, [fiscal_year_id])
```
jniles: Where is the `catch()` statement for promise?  Is there any chance that an error could occur?

server/controllers/budget.js
```diff
@@ -0,0 +1,142 @@
...
+		period         = JSON.parse(req.body.period);
+
+	if (file && file.path && file.name) {
+		var fileName = uploadDir + file.name.replace(/ /g, '_');
+
+		// File name can be more readable
+		fs.rename(file.path, fileName, function (err) {
+			if (err) { throw err; }
+			fs.stat(fileName, function (err, stat) {
+				if (err) { throw err; }
```
jniles: This should be `return next(err);`  (and the line above it as well).

jniles:
LGTM.  Thanks for going through all the revisions.

PR #778, opened 2015-10-01
Fixing parent account

mbayopanda:
In this PR, we have wrote a script `fixingParentAccount.js` for fixing parent account, some account has like parent an account_number instead of an id of the account.

The script is in `/utilities/account/fixing_parent`, 

To run the script : 
```
cd /utilities/account/fixing_parent/
node fixingParentAccount.js
```
The process is finished when you see `Update parent with success...`

### DIFF
--- a/server/models/development/update/synt.sql
+++ b/server/models/development/update/synt.sql
@@ -281,3 +281,10 @@ DROP FOREIGN KEY  `service_ibfk_1` ;
 
 INSERT INTO unit (`id`, `name`, `key`, `description`, `parent`, `url`, `path`)
 VALUES (120, 'Rapport situation group debiteur', 'TREE.DEBITOR_GROUP_REPORT', 'pour voir le rapport pdf detaille de group de debiteur', 10, '/partials/reports/debitor_group_report', '/reports/debitor_group_report/');
+
+-- Update account parent to account.is instead of account_number
+-- Date: 2015-10-01
+-- By: Bruce M.
+
+UPDATE account a JOIN account b ON a.parent=b.account_number 
+SET a.parent=b.id;
--- a/server/models/development/update/synt.sql
+++ b/server/models/development/update/synt.sql
@@ -281,10 +281,3 @@ DROP FOREIGN KEY  `service_ibfk_1` ;
 
 INSERT INTO unit (`id`, `name`, `key`, `description`, `parent`, `url`, `path`)
 VALUES (120, 'Rapport situation group debiteur', 'TREE.DEBITOR_GROUP_REPORT', 'pour voir le rapport pdf detaille de group de debiteur', 10, '/partials/reports/debitor_group_report', '/reports/debitor_group_report/');
-
--- Update account parent to account.is instead of account_number
--- Date: 2015-10-01
--- By: Bruce M.
-
-UPDATE account a JOIN account b ON a.parent=b.account_number 
-SET a.parent=b.id;
--- a/utilities/account/fixing_parent/fixingAccountParent.js
+++ b/utilities/account/fixing_parent/fixingAccountParent.js
@@ -14,6 +14,7 @@ db.initialise();
 	.then(executeAccount)
 	.then(unmatchAccount)
 	.then(updateParent)
+	.then(success)
 	.catch(error)
 	.done(endProcess);
 
@@ -82,17 +83,18 @@ db.initialise();
 		/*
 		 * Objectif : Application des mise a jour dans la base de donnees
 		 */
-		var dbPromises = [];
-		accounts.forEach(function (acc) {
-			var sql = 'UPDATE account SET parent=? WHERE account.id=? ;';
-			var value = [acc.parent, acc.id];
-			dbPromises.push(db.exec(sql, value));
+		var sql = 'UPDATE account SET parent=? WHERE account.id=? ;';
+		var dbPromises = accounts.map(function (acc) {
+			return db.exec(sql, [acc.parent, acc.id]);
 		});
 		return q.all(dbPromises);
 	}
 
-	function endProcess() {
+	function success() {
 		console.log('Update parent with success...');
+	}
+
+	function endProcess() {
 		process.exit(0);
 	}
 
--- a/utilities/account/fixing_parent/fixingAccountParent.js
+++ b/utilities/account/fixing_parent/fixingAccountParent.js
@@ -1,0 +1,96 @@
+// Fixing account parent
+'use strict';
+
+var db = require('../../../server/lib/db'),
+		q  = require('q');
+
+db.initialise();
+
+(function queryAccount() {
+	var accounts;
+	var sqlUpdateMatch = 'UPDATE account a JOIN account b ON a.parent=b.account_number SET a.parent=b.id ;';
+	var sqlAccount 		 = 'SELECT * FROM account;';
+
+	executeUpdate()
+	.then(executeAccount())
+	.then(unmatchAccount)
+	.then(updateParent);
+
+	function executeUpdate() {
+		/*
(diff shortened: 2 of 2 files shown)


utilities/account/fixing_parent/fixingAccountParent.js
```diff
@@ -0,0 +1,96 @@
...
+
+db.initialise();
+
+(function queryAccount() {
+	var accounts;
+	var sqlUpdateMatch = 'UPDATE account a JOIN account b ON a.parent=b.account_number SET a.parent=b.id ;';
+	var sqlAccount 		 = 'SELECT * FROM account;';
+
+	executeUpdate()
+	.then(executeAccount())
```
jniles: I think you wanted to write `.then(executeAccount)`.  Writing `.then(executeAccount())` will execute the function before the original promises was resolved!

utilities/account/fixing_parent/fixingAccountParent.js
```diff
@@ -0,0 +1,96 @@
...
+		accounts.forEach(function (acc) {
+			if (String(acc.account_number).indexOf(acc.parent) === 0) {
+				// account.parent fait partiellemnt partie de account.account_number
+				var chaine = String(acc.parent);
+				var end    = false;
+				var cut    = 0;
+				var accountId;
+				// on retranche progressivement un caractere de account.parent
+				// afin de trouver le bon compte
+				while (!end && cut !== chaine.length) {
```
jniles: In this loop, `end` acts like a `break;` statement.  This might be better written as:
```js
while (cut !== chaine.length) {  // this could even be cut <= chaine.length
  accountId = getAccountId(chaine.substr(0, chaine.length - cut));
  if (accountId) {
    acc.parent = accountId;
    break;
  }
  cut++;   // no need to check else, because we broke out of the loop before here!
}
```

utilities/account/fixing_parent/fixingAccountParent.js
```diff
@@ -0,0 +1,96 @@
+// Fixing account parent
+'use strict';
+
+var db = require('../../../server/lib/db'),
+		q  = require('q');
+
+db.initialise();
+
+(function queryAccount() {
+	var accounts;
```
jniles: The local variable `accounts` is unneeded.  You can keep the accounts data within the promise chain.  Here's how:
```js
executeUpdate()
.then(executeAccount) // executeAccount returns a list of accounts, so I'll pass that list to the next function
.then(unmatchAccount) // unmatchAccount receives a list of accounts from execute account, and returns the same list at the end of execution.
.then(updateParent);  // updateParent receives a list of accounts from unmatchAccount()
```

Now, you just have to change the function signature of unmatchParent to this:
```
function unmatchParent(accounts) {
  /* code */
  return accounts;
}
```
Also, you have to change updateParent's signature.
```js
function updateParent(accounts) {
  /* code */
}
```

utilities/account/fixing_parent/fixingAccountParent.js
```diff
@@ -0,0 +1,96 @@
...
+	var sqlUpdateMatch = 'UPDATE account a JOIN account b ON a.parent=b.account_number SET a.parent=b.id ;';
+	var sqlAccount 		 = 'SELECT * FROM account;';
+
+	executeUpdate()
+	.then(executeAccount())
+	.then(unmatchAccount)
+	.then(updateParent);
```
jniles: What about a catch statement?  You should at least let the user know that something went wrong!

(also, you should always end a promise chain with a `.done()` method.

server/models/development/update/synt.sql
```diff
@@ -281,3 +281,10 @@ DROP FOREIGN KEY  `service_ibfk_1` ;
 
 INSERT INTO unit (`id`, `name`, `key`, `description`, `parent`, `url`, `path`)
 VALUES (120, 'Rapport situation group debiteur', 'TREE.DEBITOR_GROUP_REPORT', 'pour voir le rapport pdf detaille de group de debiteur', 10, '/partials/reports/debitor_group_report', '/reports/debitor_group_report/');
+
```
DedrickEnc: As discussed, this line can be removed because the script handle this situation already.

utilities/account/fixing_parent/fixingAccountParent.js
```diff
@@ -0,0 +1,96 @@
...
+		/*
+		 * Objectif : Application des mise a jour dans la base de donnees
+		 */
+		var dbPromises = [];
+		accounts.forEach(function (acc) {
+			var sql = 'UPDATE account SET parent=? WHERE account.id=? ;';
+			var value = [acc.parent, acc.id];
+			dbPromises.push(db.exec(sql, value));
+		});
+		q.all(dbPromises)
```
jniles: To make this code really modular, you might want to do this
```js
function updateParent(accounts) {
  var sql = 'UPDATE account SET parent = ? WHERE account.id = ?;';  // this is a template, so we do not need to define it with every single execution of the loop!

  var promises = accounts.map(function (account) {
    return db.exec(sql, [account.parent, account.id]);
  });

  return q.all(promises);
}
```
Now, you can end your promise chain (from the beginning of this closure, line 17) with a success message!
```js
/* code */
.then(unmatchParent)
.then(updateParent)  // this returns q.all()
.then(function () {
  console.log('Update Success...');
})
.catch(function () {
  console.log('Update Failure...'):
})
.done();
```

jniles:
This code works, and is reasonably clear.  The comments are very well done.

If you have time to make the suggested improvements, they will make the code more readable, and possibly clearer to the user (with an error message, for example!).  It should also be good practice working with promise chains.

utilities/account/fixing_parent/fixingAccountParent.js
```diff
@@ -0,0 +1,102 @@
...
+
+		return accounts;
+	}
+
+	function updateParent(accounts) {
+		/*
+		 * Objectif : Application des mise a jour dans la base de donnees
+		 */
+		var dbPromises = [];
+		accounts.forEach(function (acc) {
```
jniles: There are two observations to make this code clearer:
 - `var sql` is not dependent upon the variable `acc`, but you create a new `var sql` for each item in the loop.  This is wasteful (lots of allocating and de-allocating, unless the compiler is smart), and makes the programmer think that `var sql` might somehow be different on each iteration of the loop.  It is better to move this out of the loop!
```js
var sql = 'UPDATE ... ;';
accounts.forEach(function (acc) { /* loop */ });
```
 - Since you are creating a new array, it is _much_ clearer to use [Array.map]().  The `map()` function immediately tells the programmer that you are transforming each item in the array into one output.  In this case, you are transforming the array of account definitions into an array of UPDATE statements. Like this:
```js
var sql = 'UPDATE ... ;';
var promises = accounts.map(function (account) {
  return db.exec(sql, [account.parent, account.id]); // no need for temporary variables, impossible for local variables to change the global state.
});
return q.all(promises);
```

You could even combine this into a single statement, but I think it is more difficult to read.
```js
var sql = 'UPDATE ... ;';
return q.all(accounts.map(function (account) { /* code */ }));
```

utilities/account/fixing_parent/fixingAccountParent.js
```diff
@@ -0,0 +1,102 @@
...
+			var sql = 'UPDATE account SET parent=? WHERE account.id=? ;';
+			var value = [acc.parent, acc.id];
+			dbPromises.push(db.exec(sql, value));
+		});
+		return q.all(dbPromises);
+	}
+
+	function endProcess() {
+		console.log('Update parent with success...');
```
jniles: `endProcess()` is called within the `done()` callback, which will be called with success __or with error__.  You should put the console.log() statement in a `.then()` callback which will only be called on success.

jniles:
This code looks really good now.  Thanks!

PR #780, opened 2015-10-02
FIX(find-patient): remove SOUNDEX algorithm

### LINKED ISSUE #708: Stop using SOUNDEX() for find-patient directive
After a couple weeks of trying to locate certain patients, I say we scrap this idea.  I've run into several problems, which are the following:

 1. MySQL's [SOUNDEX()](http://www.databasejournal.com/features/mysql/mysql-fuzzy-text-searching-using-the-soundex-function.html) is for American English.  The error rate is very high on our current implementation - we do not eliminate patients fast enough with more keystrokes.
 2. When I put in the SOUNDEX() algorithm, I thought we were only looking at two fields, first name and last name.  This yields two (or four, depending on how you want to look at it) possible orderings.  Later, we started looking at (optionally) three fields, which yields nine possible orderings.  This introduces *far* too many errors.  Since we are matching against three fields, let's just go back to strict equality.
 3. The typeahead's highlights don't really match up with SOUNDEX matches.  (This could be worked on, but, why put in the effort?)
 
In my view we have two options.
 1. Go back to strict matching.  This is the easiest, and probably makes the most sense from a user perspective.  No additional work is required, we can just reverse the changes I made a while ago.
 2. Implement a more robust algorithm ([Levenshtein Distance](https://en.wikipedia.org/wiki/Levenshtein_distance)).  This is a bad idea from a maintainability standpoint because IMA's development team will have to know something about algorithms to understand the code.  Also, this logic would probably be done in MySQL for performance reasons.  We have yet to discuss how to install MySQL functions, and it's against bhima's current practice to have MySQL functions.

jniles:
I've updated the patient find directive to use strict matching on
first_name, last_name, middle_name.  We do all possible combinations of
matches (9 total) to ensure that we do not miss an data inserted in
incorrect order.

This commit also updates the find-patient directive to meet the styleguide
standards.

Fixes #708.

### DIFF
--- a/client/src/js/directives/findpatient.js
+++ b/client/src/js/directives/findpatient.js
@@ -1,5 +1,20 @@
 angular.module('bhima.directives')
-.directive('findPatient', ['$compile', '$http', 'validate', 'messenger', 'appcache', function($compile, $http, validate, messenger, AppCache) {
+.directive('findPatient', FindPatientDirective);
+
+FindPatientDirective.$inject = [
+'$compile', '$http', 'validate', 'messenger', 'appcache'
+];
+
+/**
+* The FindPatient directive allows a user to serach for a patient by either the
+* patient identifier (Project Abbreviation + Reference) or by typeahead on patient
+* name.
+*
+* The typeahead loads data as your type into the input box, pinging th URL
+* /patient/search/fuzzy (for legacy reasons, not actually a fuzzy search).  The
+* HTTP endpoints sends back 10 results which are then presented to the user.
+*/
+function FindPatientDirective($compile, $http, validate, messenger, AppCache) {
   return {
     restrict: 'AE',
     templateUrl : 'partials/templates/findpatient.tmpl.html',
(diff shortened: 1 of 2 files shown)


client/src/js/directives/findpatient.js
```diff
@@ -1,5 +1,20 @@
...
+* name.
+*
+* The typeahead loads data as your type into the input box, pinging th URL
+* /patient/search/fuzzy (for legacy reasons, not actually a fuzzy search).  The
+* HTTP endpoints sends back 10 results which are then presented to the user.
+*/
+function FindPatientDirective($compile, $http, validate, messenger, AppCache) {
   return {
     restrict: 'AE',
     templateUrl : 'partials/templates/findpatient.tmpl.html',
```
DedrickEnc:
```
function searchReference(ref) {
        var url = '/patient/search/reference/';
        $http.get(url + ref)
        .success(selectPatient)
        .error(function (err) {
          console.error(err); 
        })
        .finally();
      }
```
As discussed, the **then, catch** is more recommanded because .success will be deprecated.

jniles:
@DedrickEnc the proposed changes have been made.

PR #781, opened 2015-10-05
FIX(travis): Make travis successfully build project

jniles:
We now build the database within the .travis.yml file using travis's
root user on MySQL.  Unfortunately, since our build process is not well
established, we have to hard-code the user and password into the
.travis.yml.  See this file for suggestions on how to improve our build
process.

Also included: before running, we install gulp & bower globally and run bower
install.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,18 +1,32 @@
 sudo: false
 language: node_js
 node_js:
-  - "stable"
+  - "node"
+  - "iojs"
+  - "4.1"
+  - "4.0"
   - "0.12"
   - "0.11"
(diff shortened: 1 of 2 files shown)


jniles:
@IMA-WorldHealth/owners, I want to you be aware of the changes here.  I think it's important to get our testing back online, so I've merged this directly into the repository; however, it hard-codes the bhima user details into the build script (which is probably bad practice).

What do you think about having a `build.sh` script that setups up lots of these details?  Then Travis can simply use `sh build.sh` to setup and install the users, database, and dependencies?

PR #782, opened 2015-10-05
Perf(general ledger report): Faster PDF rendering

jniles:
I've increased the rendering speed of the PDF report by migrating all
the sorting and grouping logic from JS to SQL.  This results in much
cleaner code (if you read SQL).

Additional fixes include:
 - number formatting for debit/credit fields
 - date now established at request time, rather than at first run time

NOTE
Trying to print the entire general ledger still results in a broken
(timed out) HTTP request.  We need to disallow printing a max number of
transactions/rows/dates to avoid needlessly hanging the server.


DedrickEnc:
Yes, as discussed.

PR #784, opened 2015-10-06
FIX(balance sheet): balance sheet rewrite

jniles:
I've updated the balance sheet PDF report to filter for ohada
accounts.  The parent-child relationship has been made clearer, and
numeral formatting is applied to every line of the balance sheet.

### DIFF
--- a/server/controllers/reports_proposed/data/balance_sheet.js
+++ b/server/controllers/reports_proposed/data/balance_sheet.js
@@ -18,7 +18,7 @@ var balanceDate = new Date();
 
 // This method builds a tree data structure of
 // accounts and children of a specified parentId.
-function getChildren(accounts, parentId, depth) {
+function buildAccountTree(accounts, parentId, depth) {
   var children;
 
   // Base case: There are no child accounts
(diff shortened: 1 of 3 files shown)


server/controllers/reports_proposed/data/balance_sheet.js
```diff
@@ -10,10 +16,6 @@ var ROOT_ACCOUNT_ID = 0;
 var formatDollar = '$0,0.00';
 var balanceDate = new Date();
 
-// TODO Query for balance and title account IDs
-var balanceAccountId = 2;
-var titleAccountId = 3;
-
 // This method builds a tree data structure of
 // accounts and children of a specified parentId.
 function getChildren(accounts, parentId, depth) {
```
DedrickEnc: The name buildAccountTree is best then getChildren for this function

server/controllers/reports_proposed/data/balance_sheet.js
```diff
@@ -1,5 +1,11 @@
-// reports_proposed/data/balance_sheet.js
```
DedrickEnc: This report should include every account, even income expense, it is called a provisional balance, because it is generated before closing the fiscal year.

jniles: What you are asking for is not a balance sheet, so it doesn't apply to this PR.  This PR fixes/creates a balance sheet which filters only on balance accounts.

server/controllers/reports_proposed/data/balance_sheet.js
jniles: Sure, buildAccountTree is a good name.  I'll fix this in a bit.

jniles:
@DedrickEnc, the proposed changes have been made.

PR #790, opened 2015-10-07
FIX(trialbalance): prevent posting unbalanced transactions

### LINKED ISSUE #785: Trial Balance Accepts an Imbalanced Set of Transactions
The trial balance may not report an error for a set of imbalanced transactions.  Here's photographic proof:

trialbalancebug

It is unclear to me where this might be taking place, but this is definitely a bug.

jniles:
This is a critical fix where the server would never report unbalanced
transactions resulting from a typo in the code.  This has been fixed.

Closes #785.

### DIFF
--- a/server/controllers/trialbalance.js
+++ b/server/controllers/trialbalance.js
@@ -65,12 +65,12 @@ function aggregate(property, array) {
 
 // creates an error report for a given code
 function createErrorReport(code, isFatal, rows) {
-  return q({
+  return {
     code : code,
     fatal : isFatal,
     transactions : rows.map(function (row) { return row.trans_id; }),
     affectedRows : aggregate('count', rows)
-  });
+  };
 }
 
 
@@ -89,7 +89,7 @@ function checkAccountsLocked(transactions) {
     // if nothing is returned, skip error report
     if (!rows.length) { return; }
 
-    // returns a promise error report
+    // returns a error report
     return createErrorReport('ERROR.ERR_LOCKED_ACCOUNTS', true, rows);
   });
 }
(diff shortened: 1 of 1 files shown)


server/controllers/trialbalance.js
```diff
@@ -187,7 +187,7 @@ function checkDebtorCreditorExists(transactions) {
     // if nothing is returned, skip error report
     if (!rows.length) { return; }
 
-    // returns a promise error report
+    // returns a error report
     return createErrorReport('ERROR.ERR_MISSING_DEBTOR_CREDITOR', true, rows);
   });
 }
```
DedrickEnc:
```
function checkDocumentNumberExists(transactions) {
   var sql =
     'SELECT COUNT(pj.uuid) AS count, pj.doc_num, pj.trans_id, pj.deb_cred_uuid FROM posting_journal AS pj ' +
     'WHERE pj.trans_id IN (?) AND (pj.deb_cred_type = \'D\' OR pj.deb_cred_type = \'C\') ' +
     'GROUP BY pj.trans_id HAVING pj.doc_num IS NULL;';
 
   return db.exec(sql, [transactions])
   .then(function (rows) {
 
     // if nothing is returned, skip error report
     if (!rows.length) { return; }
 
-    // returns a promise error report
+    // returns a error report
     return createErrorReport('ERROR.WARN_MISSING_DOCUMENT_ID', false, rows);
   });
 }
```
You are checking the doc_num only in transaction which have the deb_cred_type, you should do it for each transaction.

jniles: For something like an internal transfer (gain or loss on exchange, for example), an electronic transfer (such as a bank transfer), or a balancing transaction, it will not have a document number.  So it really only makes sense to refer to documents that concern two different parties, such as the hospital and a debtor/creditor.

That being said, we should figure out when we want to use a inv_po_id versus a doc_num.  I have a feeling that these should be one field, rather than two.

DedrickEnc: I agreee, about the first paragraph.
For the second, that should be discussed.

jniles: :+1:

PR #791, opened 2015-10-07
Reorder unit and new reports folders

mbayopanda:
In this PR, we have reorder the side menu of BHIMA and create folders units for reports


jniles:
LGTM.

PR #792, opened 2015-10-08
Fixing BHIMA menu names

mbayopanda:
In this PR, we have fixed some names of BHIMA side menu. Menu Names doesn't have to be too long


jniles:
LGTM.

client/src/i18n/en.json
```diff
@@ -2939,7 +2938,7 @@
    "PAYMENT_PERIOD"            : "Payment Period Management",
    "PAYROLL"                   : "Payroll",
    "PAYROLL_REPORT"            : "Payroll",
-   "PERMISSION"                : "Users and Permissions",
+   "PERMISSION"                : "Users",
```
jniles: Why is this no longer Users and Permissions?  Why does the key PERMISSION map to "Users"?  This is very confusing.

PR #793, opened 2015-10-08
Fixing users and permissions name

mbayopanda:
In this PR, we have fixed the name of users and permissions module


jniles:
Ahh, so much more meaningful.

PR #795, opened 2015-10-09
Payslip employee with directive

mbayopanda:
In this PR we have used Angular Directive for the "Multi Payroll" ui

### DIFF
--- a/client/src/js/directives/payslip_employee.js
+++ b/client/src/js/directives/payslip_employee.js
@@ -1,0 +1,19 @@
+/**
+  * This Directive is responsible for generate an employee block of data
+  */
+angular.module('bhima.directives')
+.directive('bhimaPayslipEmployee', function () {
+  return {
+    restrict    : 'E',
+    templateUrl : '/partials/templates/payslipEmployee.tmpl.html',
+    scope       : {
+      row             : '=employee',
+      cotisations     : '=cotisations',
+      rubriques       : '=rubriques',
+      taxesEmployee   : '=taxesEmployee',
+      taxesEnterprise : '=taxesEnterprise',
+      currency        : '=currency',
+      submit          : '&onSubmit'
+    }
+  };
+});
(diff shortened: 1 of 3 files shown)


client/src/js/directives/payslip_employee.js
```diff
@@ -0,0 +1,19 @@
+/**
+  * This Directive is responsible for generate an employee block of data
+  */
+angular.module('bhima.directives')
+.directive('bhimaPayslipEmployee', function () {
+  return {
+    restrict    : 'E',
+    templateUrl : '/partials/templates/payslipEmployee.tmpl.html',
+    scope       : {
+      row             : '=employee',
```
jniles: Why not call this `employee`?  It would be something like `scope : { employee : '=' }`.

client/src/partials/templates/payslipEmployee.tmpl.html
```diff
@@ -0,0 +1,148 @@
+<div class="panel panel-primary">
+  <div ng-click="row.visible=!row.visible" class="panel-heading">
```
jniles: This should be an `isVisible` attribute on the directive scope, since `employee.visible` doesn't really make a lot of sense.

jniles:
This is a good start.

I noticed you do not create a link function or controller.  It would be nice to encapsulate the HTTP request to the server within this directive.  That way the parent controller does not have to worry about the details of engaging an employee, it can just download the correct data, create all the employee payslip directives, and concern itself with changing the currency.

The directive should have a controller with a `submit()` method which sends the HTTP request to approve payment for the employee.  It should perform any data validation necessary before sending the HTTP request.  If the request succeeds, the directive should call the `onSubmit()` method that was passed to it with `true` as the only parameter.  If the HTTP request fails, call `onSubmit(false)`.

jniles:
After some discussion, it doesn't seem like a directive is the appropriate way forward for this module, unless we rethink how payroll happens.  Thanks for the effort though!

PR #800, opened 2015-10-12
Rewrite Group Invoice

### LINKED ISSUE #674: Tiers Support  (group_invoice) shows UUIDs instead of patient IDs
The `group_invoice` module shows very ugly UUIDs instead of the patient's hospital ID.  This is shown below:
uuids

We should probably show human readable ids (e.g `HBB1`) instead.

#### Edit
There are lots of issues with this page, I'll list them here for reference
 1. The title says "Support for Debtor Debtor" -- careless mistake, we can remove the first "Debtor"
 2. Invoice ID is also a UUID.
 3. Balance is confusing -- it is the _remaining_ balance after the proposed transaction.  Maybe we should have two columns: "remaining balance" and "transaction balance"?
 4. There is lots and lots of unnecessary whitespace
 5. The debtor button in the top util bar doesn't seem to do anything, we should remove it.

jniles:
This PR rewrites the client-side code base of Group Invoice (`/group_invoice`), but does not change the logic.  The following changes have been made:
 - No more UUIDs in the view
 - All currency filters reference the sale currency ID
 - Complete removal of `$scope` in favor of `controller as` syntax
 - Loading indicators while data is fetched from the server
 - Fixes to broken HTML.

Fixes #674.

### DIFF
--- a/client/src/partials/group_invoice/group_invoice.js
+++ b/client/src/partials/group_invoice/group_invoice.js
@@ -1,24 +1,38 @@
 angular.module('bhima.controllers')
 .controller('GroupInvoiceController', GroupInvoiceController);
 
-
 GroupInvoiceController.$inject = [
-  '$scope', '$translate', 'connect', 'validate', 'messenger', 'uuid', 'SessionService',
+  '$translate', 'connect', 'validate', 'messenger', 'uuid', 'SessionService',
 ];
 
-function GroupInvoiceController($scope, $translate, connect, validate, messenger, uuid, SessionService) {
+/**
+* Group Invoice Controller
+*
+* This controller allows conventions to pay for bills of patients.
+*/
+function GroupInvoiceController($translate, connect, validate, messenger, uuid, SessionService) {
   var vm = this;
 
   var dependencies = {};
-  $scope.action = '';
-  $scope.convention = '';
-  $scope.selected = {};
-  $scope.paying = [];
-  $scope.loading = false;
+  vm.action = 'default';
+  vm.loading = false;
+  vm.queue = [];
 
-  // get enterprise
-  $scope.project = SessionService.project;
-  $scope.enterprise = SessionService.enterprise;
+  // import session variables
+  vm.project = SessionService.project;
+  vm.enterprise = SessionService.enterprise;
+
+  // bind methods
+  vm.authorize = authorize;
+  vm.setDebtor = setDebtor;
+  vm.examineInvoice = examineInvoice;
+  vm.enqueue = enqueue;
+  vm.dequeue = dequeue;
+  vm.pay = pay;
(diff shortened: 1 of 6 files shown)


client/src/partials/group_invoice/group_invoice.js
```diff
@@ -1,193 +1,217 @@
...
+
+GroupInvoiceController.$inject = [
+  '$scope', '$translate', 'connect', 'validate', 'messenger', 'uuid', 'SessionService',
+];
+
+function GroupInvoiceController($scope, $translate, connect, validate, messenger, uuid, SessionService) {
+  var vm = this;
+
+  var dependencies = {};
+  $scope.action = '';
```
DedrickEnc: You still using ```$scope```, every information to expose must be attached to the ```vm```.
You don't need $scope here.

client/src/partials/group_invoice/group_invoice.js
```diff
@@ -1,193 +1,217 @@
...
+    var payment = $scope.payment;
+    payment.project_id = $scope.project.id;
+    payment.group_uuid = $scope.selected.convention.uuid;
+    payment.debitor_uuid  = $scope.selected.debitor.uuid;
+    payment.total = $scope.paymentBalance;
+    payment.date = new Date().toISOString().slice(0,10);
+    $scope.action = 'confirm';
+  };
+
+  $scope.$watch('paying', function () {
```
DedrickEnc: Only paying can be attached to both vm and $scope like ```$scope.paying = vm.paying = object``` for watch considerations.

jniles: Actually, you don't even need to use `$scope` for this as well.   This was a poor design, and never should have used a `$watch`.

I wasn't going to do the full code rewrite, since the issue (#767) only concerned itself with the UI.  Would you prefer that I rewrite this module?

DedrickEnc: Yes you are write, many articles advise to avoid the using of ```$scope.$watch```

jniles: More than that, you don't actually need `$scope.$watch`, since all changes are triggered in the controller!  `$scope.$watch` might be useful if a user makes an asynchronous change (such as typing in a text box), but this was just fired on data load, as far as I can tell.

That article you linked is good, although I would recommend _never_ using `$setInterval()` to do the same thing as watch.  Instead, just design your system so that you don't need a `$watch()`.

Should I do a full rewrite to remove `$scope`?

jniles: Hmm. Okay, guess I'll do a rewrite.

jniles:
@DedrickEnc , I've rewritten the module to no longer use `$scope`.

PR #803, opened 2015-10-13
FIX(mysql pool): set higher handshake timeout

jniles:
Node v4.2.0 seems to result in database build errors due to handshake
timeouts.  This commit raises the timeout limit for the connection pool
to allow Travis CI to build the database correctly.


jniles:
Hmm..  No luck yet.

Issue #709, opened 2015-09-08
Redesign ExchangeService


jniles:
I do not have a solid design proposal yet, but given the work @sfount has done on currency, we owe it to ourselves to redesign our exchange service.  There are three issues I can think of:
 1.  We are not guaranteed to have an exchange rate on application startup.  Just like with the `SessionService` we should be guaranteed to know if we have an exchange rate or not for today when we use the ExchangeService.
 2.  We shouldn't enforce a daily exchange rate.  The application should just use the newest exchange rate.  If an institution wants to make one every day, they are free to do so.
 3. The API for exchange is convoluted. [Look at the code!](https://github.com/IMA-WorldHealth/bhima/blob/master/client/src/js/services/exchange.js) We have two methods, exchange and conventir...  nevermind the fact that they are in two different languages.  This should be cleaned up immensely.

jniles:
Update: #809 addressed the third point made here, but the other two still need to be tackled.  See the Google Drive docs for an `ExchangeService` API proposal.

jniles:
This has landed in development.

PR #809, opened 2015-10-14
FIX(exchange rate): rewrite exchange controller

### LINKED ISSUE #588: Exchange rate module does not display date properly
Under set today's exchange rate in the exchange rate module, the date is blank in Firefox and 'mm/dd/yyyy' in chrome (with en_US locate).

The module works fine, but it's confusing to not know what date your are adding for.  Furthermore, the lack of appearance in Firefox is a bug.

### LINKED ISSUE #673: Set Exchange Rate Console Errors
Randomly, setting an exchange rate results in the following console error:
```js
Error: $scope.globalRates is undefined
$scope.submit/<@https://localhost:8080/js/bhima.min.js:3509:1
```

It doesn't seem to happen regularly, and it does not affect the end result (an exchange rate is added), but it should be fixed nonetheless.

jniles:
This commit rewrites the exchange controller.  The following changes
have been introduced:
 1. Simpler date changes and validation in the controller code
 2. Use AngularJS form validation
 3. Dates render properly in the client (fixes #588)
 4. Fixes #673.

### DIFF
--- a/client/src/partials/exchange_rate/exchange_rate.js
+++ b/client/src/partials/exchange_rate/exchange_rate.js
@@ -1,114 +1,125 @@
 angular.module('bhima.controllers')
-.controller('exchangeRate', [
-  '$scope',
-  'connect',
-  'messenger',
-  'appstate',
-  'validate',
-  'exchange',
-  'util',
-  function ($scope, connect, messenger, appstate, validate, exchange, util) {
-    var dependencies = {};
-    var session = $scope.session = {};
+.controller('ExchangeRateController', ExchangeRateController);
 
-    dependencies.currency = {
-      required : true,
-      query : {
-        tables : {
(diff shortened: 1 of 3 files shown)


client/src/partials/exchange_rate/exchange_rate.js
```diff
@@ -1,114 +1,125 @@
...
       }
-    };
+    }
+  };
 
-    $scope.maxDate = util.htmlDate(new Date());
-    $scope.currentDate = util.htmlDate(new Date());
-    session.is_oldDate = false;
-    session.oldDate = new Date();
+  //
```
DedrickEnc: Why not to create a data loader service to load exchange rate for this module? instead of appstate?

jniles: Your suggestion is essentially #709.  This PR is only a bug fix to address the bugs in #588 and #673.

PR #812, opened 2015-10-14
Posting Journal Refactoring

DedrickEnc:
This PR is for the posting journal refactoring, the goal was to decrease the number of controller and to prioritize services using.
So after the refactoring, I created services to do : 
1. loading data
2. handling grid
3. handling the dataview
4. handling session managent
5. handling validation
I will probably create an other service for ```Editor```
We have now three controllers : ```PostingJournalController, JournalUtilitieController``` and ```Journal.print```(not refactored yet, may be it will be transformed to a service).

It is very probable to find some mistakes that why the review is necessary also.


client/src/js/services/journalDataLoaderService.js
```diff
@@ -0,0 +1,112 @@
...
+    dataLoaderService.getAccountStore = function getAccountStore (){
+      return dataLoaderService.account;
+    };
+
+    function loadAll (){
+
+      validate.process(dependencies)
+        .then(function (data){
+          angular.extend(dataLoaderService, data);
+          df.resolve(data);
```
jniles: This code is equivalent and does not require creating a new `defer()`.

```js
function loadAll() {
  return validate.process(dependencies)
  .then(function (data) {
    angular.extend(dataLoaderService, data);
    return data;
  });
}
```

client/src/js/services/journalDataviewService.js
```diff
@@ -0,0 +1,205 @@
...
+
+      if (groupMap[targetGroup]) { groupMap[targetGroup](); }
+    };
+
+    dataviewService.setFilter = function setFilter(filter, param){ 
+      if(!filter) {throw 'filter function undefined';}
+      dataviewService.dataview.beginUpdate();
+      dataviewService.dataview.setFilter(filter);
+      dataviewService.dataview.setFilterArgs({
+        param : param
```
jniles: Is there a regular expression missing here?

client/src/js/services/journalGridService.js
```diff
@@ -0,0 +1,338 @@
...
+      if (record.cc && !Number.isNaN(Number(record.cc))) { data.cc_id = record.cc; }
+      if (record.pc && !Number.isNaN(Number(record.pc))) { data.pc_id = record.pc; }
+
+      // FIXME : Hack to get deletion to work with parser.js
+      // This is probably pretty damn insecure
+      if (record.cc === 'null') { data.cc_id = null; }
+      if (record.pc === 'null') { data.pc_id = null; }
+
+      // FIXME : Review this decision
+      data.project_id = sessionService.project.id;
```
jniles: We should probably create a project editor, where you can specify the project.

client/src/js/services/journalGridService.js
```diff
@@ -0,0 +1,338 @@
...
+      return data;
+    }
+
+    function sort (a,b) {
+      var x = a[this.sortColumn];
+      var y = b[this.sortColumn];
+      return x > y ? 1 : -1;
+    }
+
+    function doParsing (o) { return JSON.parse(JSON.stringify(o)); }
```
jniles: This should probably be a more meaning name.  This is a hack to deep-copy a JSON object, without copying any of its references.

I'd call it `clone()` or `copy` or `deepClone()` or something along those lines.

client/src/partials/journal/journal.utilities.js
```diff
@@ -1,383 +1,343 @@
...
+  //   modal.result.then(function () {
+  //     $location.path('/reports/ledger/general_ledger');
+  //   });
+  // };
+
+
+};
+
+journalUtilitiesController.$inject = ['$scope', '$translate', '$location', '$modal', 'appcache', 'connect', 'validate', 'appstate', 'messenger',  'JournalDataviewService', 'JournalColumnsService', 'JournalGridService', 'JournalDataLoaderService', 'JournalManagerService'];
+angular.module('bhima.controllers').controller('journal.utilities', journalUtilitiesController);
```
jniles: I don't think you ever use this controller.  Can it be deleted?

client/src/partials/journal/journalUtilities.js
```diff
@@ -0,0 +1,242 @@
...
+  vm.toggleEditMode = toggleEditMode;
+  vm.refreshFilter = refreshFilter;
+  vm.updateFilter = updateFilter;
+  vm.filterBy = filterBy;
+  vm.getGroupingType = getGroupingType;
+  vm.print = print; 
+  vm.trialBalance = trialBalance;
+  
+
+  $scope.$watch('filter', function () {
```
jniles: Can this use an `ng-change()` instead of the `$watch`?

client/src/partials/journal/journalUtilities.js
```diff
@@ -0,0 +1,242 @@
...
+    updateFilter();
+  }, true);
+
+  $scope.$watch('columns', function () {
+    if (!$scope.columns) { return; }
+    var columns = $scope.columns.filter(function (column) { return column.visible; });
+   
+    cache.put('columns', columns);
```
jniles: This logic should be in the grid service.

client/src/js/services/journalDataLoaderService.js
```diff
@@ -0,0 +1,112 @@
...
+
+    dataLoaderService.loadAdditionalData = function loadAdditionalData (){
+      return validate.process(dependencies, ['debtor', 'creditor', 'invoice', 'period', 'cost_center', 'profit_center']);
+    };
+
+    dataLoaderService.loadAccountData = function loadAccountData(){
+      return validate.process(dependencies, ['account']);
+    };
+
+    dataLoaderService.SetAccountStore = function SetAccountStore (accountStore){
```
jniles: This probably should be `dataLoaderService.setAccountStore`.

client/src/js/services/journalDataLoaderService.js
```diff
@@ -0,0 +1,112 @@
...
+          }
+        }
+      }
+    };
+
+    dataLoaderService.loadJournalRecord = function loadJournalRecord (){
+      return validate.process(dependencies, ['journalRecord']);
+    };
+
+    dataLoaderService.loadAdditionalData = function loadAdditionalData (){
```
jniles: Will there ever be a case when we will want to load only journal data, without the additional data?  Why make this two different methods?

client/src/partials/journal/journalUtilities.js
```diff
@@ -0,0 +1,242 @@
+var journalUtilitiesController = function ($scope, $translate, $location, $modal, Appcache, connect, validate, appstate, messenger, dataviewService, columnsService, gridService, dataLoaderService, managerService){
+  var vm = this, cache = new Appcache('journal.utilities');
+
+  vm.columnsService = columnsService;
+  vm.gridService = gridService;
+  vm.dataviewService = dataviewService;
+  vm.managerService = managerService;
+  vm.dataLoaderService = dataLoaderService; 
+  // vm.hasData = vm.dataviewService.hasData(); //FIX ME : this line is executed before de the dataview is populated
+  $scope.filter = vm.filter = { by : {} };
```
jniles: Can all of the filtering logic take place in a FilterService?

client/src/partials/journal/postingJournalController.js
```diff
@@ -0,0 +1,354 @@
...
+    angular.extend(vm, models);
+    vm.journal = new Store({ data : vm.dataviewService.dataview.getItems() });
+    vm.dataviewService.subscribeToOnRowCountChanged(vm.grid);
+    vm.dataviewService.subscribeToOnRowsChanged(vm.grid);
+    vm.gridService.subscribeToOnCellChange();
+    vm.gridService.subscribeToOnSort();
+    vm.gridService.subscribeToOnBeforeEditCell();
+    vm.gridService.subscribeToOnclick();
+
+    var editors = {
```
jniles: The editors need to be in an EditorService, but I don't think you need to do that in this pull request.

client/src/partials/journal/postingJournalController.js
```diff
@@ -0,0 +1,354 @@
...
+  function handleErrors (error) {
+    messenger.danger('An error occured ' + JSON.stringify(error));
+  }
+
+  function initialise (models) {
+    angular.extend(vm, models);
+    vm.journal = new Store({ data : vm.dataviewService.dataview.getItems() });
+    vm.dataviewService.subscribeToOnRowCountChanged(vm.grid);
+    vm.dataviewService.subscribeToOnRowsChanged(vm.grid);
+    vm.gridService.subscribeToOnCellChange();
```
jniles: Since you are calling four methods on gridService, could you group these into a single method call?  Something like
`gridService.initGridListeners()` or `gridService.initEventHandlers()`?

client/src/partials/journal/postingJournalController.js
```diff
@@ -0,0 +1,354 @@
...
+  .catch(handleErrors);  
+  
+  function isDefined (d) { return angular.isDefined(d); }
+  function handleErrors (error) {
+    messenger.danger('An error occured ' + JSON.stringify(error));
+  }
+
+  function initialise (models) {
+    angular.extend(vm, models);
+    vm.journal = new Store({ data : vm.dataviewService.dataview.getItems() });
```
jniles: Does this need to be a Store?  I know that the code used to use a store, but I think this might be a cause for significant slowdown!  If you can easily migrate away from the store, that would be a nice performance bump.

client/src/js/services/journalGridService.js
```diff
@@ -0,0 +1,338 @@
...
+        'addRow'            : addRow,
+        'saveTransaction'   : saveTransaction
+        // 'deleteTransaction' : deleteTransaction
+      };
+      classes.forEach(function (cls) {
+        if (buttonMap[cls]) { buttonMap[cls](args); }
+      });
+    }
+
+    function editTransaction(args) {
```
jniles: Would it be possible to have a `GridTransactionService` to handle the creation of transactions?  It would take care of things like:
 1. Create the initial template
 2. TransactionService.addRow(), TransactionService.removeRow(idx)
 3. The validation logic could be contained either in the TransactionService or in a separate TransactionValidationService.

I think that would help reduce the complexity in both this controller and managerService.

client/src/js/services/journalGridService.js
```diff
@@ -0,0 +1,338 @@
+angular.module('bhima.services')
+.service('JournalGridService', ['JournalDataviewService', 'JournalColumnsService', 'JournalManagerService', 'uuid', 'SessionService', 'messenger', 'JournalValidationService', 'liberror', 'JournalDataLoaderService', 'util', 'connect', '$q', '$translate', 'precision', '$window',
```
jniles: Could you use the style we've been using so far? Like this:
```js
angular.module('some.module')
.service('SomeService', SomeService);

SomeService.$inject = [ /* dependencies */];

function SomeService(/* aliases for the dependencies */ ) {
  var service = this;
  / * code */
}
```

Using the standard name `var service = this;` will help me review this more effectively, since I'll always know what `this` is.

client/src/js/services/journalDataviewService.js
```diff
@@ -0,0 +1,205 @@
...
+
+    dataviewService.addNewItem = function addNewItem(row){
+      dataviewService.dataview.addItem(row);
+    };
+
+    dataviewService.getItem = function getItem(row){
+      return dataviewService.dataview.getItem(row);
+    };
+
+    dataviewService.getItems = function getItem(){
```
jniles: I notice that you duplicate a lot of methods from dataview --> dataviewService.  Is this easier than having a controller call `dataviewService.dataview.getItems()`? 

If you think aliasing is advantageous, I recommend you use the exact same API as dataview. So, you can write this:
```js
// alias dataview properties for faster lookups
service.getItems = service.dataview.getItems;
service.getItem = service.dataview.getItem;
service.addItem = service.dataview.addItem;
service.updateItem = service.dataview.updateItem;
// etc...
```
Avoiding changing the API will cause less confusion among programmers.  It also allows you to write compact code.  However, I'm not certain why a controller shouldn't just call `DataviewService.dataview.getItems()` anyway.

client/src/js/services/journalDataviewService.js
```diff
@@ -0,0 +1,205 @@
+angular.module('bhima.services')
+.service('JournalDataviewService', ['JournalDataLoaderService', 'JournalManagerService', '$translate',
+  function (dataLoaderService, managerService, $translate) {
+    var dataviewService = this;
+    var inlineFilter = true;
```
jniles: Is there an advantage to putting this in a variable?

client/src/js/services/journalDataviewService.js
```diff
@@ -0,0 +1,205 @@
...
+
+    dataviewService.aggregates = true;    
+    dataviewService.gimp = new Slick.Data.GroupItemMetadataProvider(); //plugin to group metadata for slick grid
```
jniles: Nice alias!

client/src/js/services/journalDataviewService.js
```diff
@@ -0,0 +1,205 @@
...
+    dataviewService.dataLoaderService = dataLoaderService;
+    dataviewService.managerService = managerService;
+    dataviewService.dataview = new Slick.Data.DataView({
+      groupItemMetadataProvider: dataviewService.gimp,
+      inlineFilter: inlineFilter
+    });
+
+
+    dataviewService.setAggregate = function setAggregate (ans){
```
jniles: Do you ever use these?  If so, would `toggleAggregate()` be a better method?

client/src/js/services/journalDataviewService.js
```diff
@@ -0,0 +1,205 @@
...
+          lazyTotalsCalculation : true
+        });
+      }
+
+      function groupByAccount () {
+        dataviewService.dataview.setGrouping({
+          getter: 'account_id',
+          formatter: function(g) {
+            var account_txt;
+            if (dataviewService.dataLoaderService.getAccountStore() && g.rows[0].account_number) {account_txt = dataviewService.dataLoaderService.getAccountStore().get(g.rows[0].account_number).account_txt || '';}
```
jniles: Man, this is a long line.  Any way you can cache some variables to make it not quite so hard to read?  I'm thinking `var accounts = dataviewService.dataLoaderService.getAccountStore()` would help, and maybe doing a bit more than that.

jniles:
@DedrickEnc, this is awesome work.  This was a difficult task and you've made significant progress.  I'm very pleased with this PR.

I've made some initial observations which should give you ideas on how the code might evolve.  I'll let you review my feedback, and make some changes.  Feel free to reply with more details if I have not understood something.  I'll review whatever changes/improvements you make and be ready to merge this in when I get back from Vanga.

jniles:
Ping me with @jniles when these edits are done.

DedrickEnc:
I will touch this PR after the trip eventually.

PR #817, opened 2015-10-19
FIX(voucher totals): Retotal voucher on recovery

### LINKED ISSUE #816: Journal Voucher : No total when reloading the form
When reloading the page to continue the transcation adding process, there is no value for total.
**Step to bug** 
1. Go to journal voucher
2. make a transaction
3. Before validate leave the page and go somewhere else
4. Return to the journal voucher page
5. Click the button to reload the previous form.

**Notice**
 No value for the sum.

jniles:
This commit adds the ability to retotal a journal voucher when the
voucher loaded from stored data.

I've also updated the controller to use $inject syntax while I was
there.  Notably, `self` --> `vm`, to standardize our controllers.

Fixes #816.

### DIFF
--- a/client/src/partials/journal_voucher/journal_voucher.js
+++ b/client/src/partials/journal_voucher/journal_voucher.js
@@ -58,10 +58,10 @@ function JournalVoucherController($scope, $http, AppCache) {
 
   // load dependencies
   $http.get('/finance/currencies')
-  .success(function (data) {
-    vm.currencies = data;
+  .then(function (response) {
+    vm.currencies = response.data;
   })
-  .error(function (error) {
+  .catch(function (error) {
     console.error(error);
   });
 
@@ -289,10 +289,10 @@ function JournalVoucherTableController($http, $q, $scope) {
 
   // load all accounts
   $http.get('/accounts?type=ohada')
-  .success(function (data) {
-    vm.accounts = data;
+  .then(function (response) {
+    vm.accounts = response.data;
   })
-  .error(handle);
+  .catch(handle);
 
   // load debtors
   $http.get('/finance/debtors')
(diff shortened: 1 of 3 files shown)


client/src/partials/journal_voucher/journal_voucher.js
```diff
@@ -1,238 +1,246 @@
...
+    rows : [] // the child
+  };
 
-    // toggle reference field
-    self.toggleReference = function () {
-      self.showReference = !self.showReference;
-    };
+  // load dependencies
+  $http.get('/finance/currencies')
+  .success(function (data) {
```
DedrickEnc: ```success``` is deprecated as discussed

jniles: This is now fixed.

client/src/partials/journal_voucher/journal_voucher.js
```diff
@@ -277,35 +289,35 @@ angular.module('bhima.controllers')
...
-  .success(function (data) {
-    self.accounts = data;
+  .then(function (response) {
+    vm.accounts = response.data;
   })
-  .error(handle);
+  .catch(handle);
 
   // load debtors
   $http.get('/finance/debtors')
```
DedrickEnc: you should use ```.then``` instead of ```.success```.

jniles: Hm.  You should refresh your page. :)

client/src/partials/journal_voucher/journal_voucher.js
```diff
@@ -370,10 +382,9 @@ angular.module('bhima.controllers')
   // set the account for a row
   // remove all debtor/creditor properties that may
   // have been set on a previous selection
-  self.setAccount = function (row) {
+  vm.setAccount = function (row) {
     row.account_id = row.account.id;
     row.deb_cred_uuid = undefined;
```
DedrickEnc: ```undefined``` means the variable is declared but no value is assigned yet, to be semantically correct you should assign ```null``` which is an assignment value, and it is also a representation of no value.

jniles:
@dedrickenc, all the fixes have landed.

PR #819, opened 2015-10-19
UI(caution): improvements to caution UI

jniles:
This commit contains significant enhancements to the caution controller
and UI.
 1. Cautions use controller as syntax
 2. AngularJS for form validation.
 3. Validation happens both in real time and at submit.  If an invalid
 form is submitted, an error message is shown.
 4. The find-patient can be refreshed.
 5. `<address>` for semantic HTML.
 6. Patient information form fields line up.
 7. Improvements to cashbox + currency lookups
 8. `.help-block` reverted to regular bootstrap behavior

### DIFF
--- a/client/src/partials/caution/caution.js
+++ b/client/src/partials/caution/caution.js
@@ -1,276 +1,252 @@
 angular.module('bhima.controllers')
-.controller('caution', [
-  '$scope',
-  '$location',
-  '$translate',
-  '$modal',
-  'validate',
-  'connect',
-  'appstate',
-  'messenger',
-  'util',
-  'uuid',
-  'appcache',
-  'exchange',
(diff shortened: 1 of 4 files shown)


client/src/partials/caution/caution.js
```diff
@@ -1,276 +1,252 @@
...
-
-    dependencies.cashboxes = {
-      query : {
-        tables : {
-          'cash_box' : {
-            columns : ['id', 'text', 'project_id']
-          }
+.controller('CautionController', CautionController);
+
+CautionController.$inject = [
```
DedrickEnc: You should delete unnecessary service such as ```appstate```

jniles: This has been fixed.

PR #825, opened 2015-10-20
FIX(exchange): set rates to today, not yesterday

### LINKED ISSUE #824: Exchange Rates are broken
PR #810 or #809 seems to have broken exchange rates.  I'm looking into this.

jniles:
This commit fixes a critical bug in which the rates were always set to
the day before.  This is do to errors in the angular validation,
stemming from differences between ngMax and the HTML max attribute.

Fixes #824.


jniles:
Hmm..  This didn't seem to fix it in the general case.  I'll take another look.

PR #829, opened 2015-10-21
Refactor/languages

sfount:
- Model languages in the database vs. using *hard coded key strings* in the settings page. 
- Standardised locale + language translation across different pages (both settings and login) 
- Propagate language models (using ids etc.) throughout language modules and caches 

I cannot see any additional places that languages or locale will have to be updated other than settings and login, the merits of wrapping these functions in a service can be discussed (although this is a relatively small job). This can either be added during merge or separately. 

*(This branch was developed some time ago, I will update it with the latest development and let someone else can merge it)*

### DIFF
--- a/server/models/development/update/synt.sql
+++ b/server/models/development/update/synt.sql
@@ -217,3 +217,22 @@ ALTER TABLE `rubric`
 CHANGE `is_advance` `is_advance` tinyint(1) DEFAULT '0';
 ALTER TABLE `rubric`  
 CHANGE `is_social_care` `is_social_care` tinyint(1) DEFAULT '0';
+
+-- @sfount
+-- 2015-15-09
+-- Model languages
+
+DROP TABLE IF EXISTS `language`;
+CREATE TABLE `language` (
+  `id` smallint(5) unsigned NOT NULL,
+  `title` varchar(35) NOT NULL,
+  `description` varchar(50),
+  `translateKey` varchar(5) NOT NULL,
+  `localeKey` varchar(5) NOT NULL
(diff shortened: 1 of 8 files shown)


server/models/development/update/synt.sql
```diff
@@ -217,3 +217,36 @@ ALTER TABLE `rubric`
...
+UPDATE `language` SET locale_key = 'en-us' WHERE id = 1;
+UPDATE `language` SET locale_key = 'fr-cd' WHERE id = 2;
+UPDATE `language` SET locale_key = 'fr-cd' WHERE id = 3;
+
+ALTER TABLE `language` MODIFY `locale_key` varchar(5) NOT NULL;
+
+-- @sfount
+-- 2015-17-09
+-- Model enterprise (application) wide default language 
+ALTER TABLE `enterprise` ADD language_id tinyint(3);
```
jniles: I remember we discussed adding languages to enterprises, but after seeing this code, would it make better sense to have languages associated with users?

sfount: Potentially - although I would go about thinking about that as a convenience. I (think?) I was thinking about it from an install perspective. 

An enterprise doesn't make sense without a currency - requiring the correct database configuration and format files etc. It also doesn't make sense to ever leave the language as something "default", especially given our context. (Another thing requiring correct databases configuration and format files). 

If we want to allow users to maintain an account across multiple computers it would be essential to *also* associate languages with users however it seems to me that this will have to default to something at user creation (enterprise language for example).

jniles:
The code here isn't voluminous or complex, I'm happy to have it merged.  If you could put that `.catch()` in and update to development, I'd be happy to pull this in.

PR #833, opened 2015-10-21
FEAT(di): require strict dependency injection

jniles:
Since we have agreed to use Controller.$inject syntax (#719), and are compiling
all our scripts through gulp, we can get some speed gains by adding the
strict DI mode directive to the main application.  For more details,
[see this link](https://docs.angularjs.org/guide/production).

I've tested most of the modules, and this change leaves them unaffected.
However, if we do find issues, we might want to add ng-annotate to our
gulp build.

### DIFF
--- a/client/src/index.html
+++ b/client/src/index.html
@@ -7,7 +7,7 @@
   <link rel="icon" href="assets/favicon.ico" type="image/x-icon">
   <link rel="stylesheet" href="css/style.min.css">
 </head>
-<body ng-app="bhima">
+<body ng-app="bhima" ng-strict-di>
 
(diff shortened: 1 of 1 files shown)


DedrickEnc:
According to angular doc, it should be used in production.
Did you try it and there is no error?

jniles:
From my pull request text:
> I've tested most of the modules, and this change leaves them unaffected.
However, if we do find issues, we might want to add ng-annotate to our
gulp build.

Yup, I tested it.  Haven't done a performance comparison, but it didn't slow anything down, as far as I can tell.

PR #837, opened 2015-10-22
Variation income expense report

DedrickEnc:
This reports show the variation between fiscal year for an income expense report.
The problem is if the previous fiscal year is locked all income expense will be balanced and the balance will be 0.
After discussion with Patrick, we discovered that, those kind of report could be generated in extra.


client/src/partials/reports_proposed/variation_exploitation/variation_exploitation.js
```diff
@@ -0,0 +1,69 @@
...
+    configurationObject.language = configuration.language.selected.value;
+    configurationObject.fy = vm.fiscal_year_id;
+    configurationObject.pfy = vm.previous_fiscal_year_id ||  vm.fiscal_year_id;
+    configurationObject.enterprise = configuration.enterprise;
+    configurationObject.project = configuration.project;
+
+    vm.building = true;
+
+    $http.post(path, configurationObject)
+      .success(function (result) {
```
jniles: Use `.then()` and `.catch()`.

client/src/partials/reports_proposed/variation_exploitation/variation_exploitation.js
```diff
@@ -0,0 +1,69 @@
...
+        // Expose generated document path to template
+        vm.building = false;
+        vm.generatedDocumentPath = result;
+      })
+      .error(function (code) {
+        vm.building = false;
```
jniles: You've written `vm.building = false;` twice, in the `.success()` and `.error()` methods.  If `vm.building` is independent of the result of the promise, you should put it in a `.finally()` method, which will by called last to clean the promise up.

client/src/partials/reports_proposed/variation_exploitation/variation_exploitation.js
```diff
@@ -0,0 +1,69 @@
...
+          columns : ['id', 'fiscal_year_txt', 'previous_fiscal_year']
+        }
+      }
+    }
+  };
+
+  vm.generate_doc = $translate.instant('VARIATION_EXPLOITATION.GENERATE_DOC');
+  vm.loading = $translate.instant('VARIATION_EXPLOITATION.LOADING');
+  vm.building = false;
+  vm.configuration = configuration;
```
jniles: It's a bit confusing to bind a series of properties to `vm` at both the top and [bottom](https://github.com/IMA-WorldHealth/bhima/pull/837/files#diff-516a7d58993316077d75d5d7e37c182fR63) of your controller function.  Could you choose one location for the bindings?

server/controllers/reports_proposed/data/variation_exploitation.js
```diff
@@ -0,0 +1,174 @@
+// reports_proposed/data/bilan.js
+// Collects and aggregates data for the enterprise bilan
+
+var q       = require('q');
```
jniles: You don't need `q` in this module.

server/controllers/reports_proposed/data/variation_exploitation.js
```diff
@@ -0,0 +1,174 @@
...
+  });
+
+  return finalList;
+}
+
+function getParentAccount(parentId){
+  var found;
+  for (var i = currentExploitations.length - 1; i >= 0; i--) {
+    if(currentExploitations[i].id == parentId){
+      found = currentExploitations[i];
```
jniles: Since you are looping from the end to the beginning, you can reduce the number of variables using a while loop.  Also, since you are doing a boolean check (which should be using triple equals `===`, by the way), you can put that in the top of the while loop and avoid the break statement.  Here's how:

```js
function getParentAccount(parentId) {
  var found = false,  // we haven't found anything yet
      i = currentExploitations.length;

  // the loop decrements our counter and checks that we haven't found anything
  while (!found && i--) {
    found = (currentExploitations[i].id === parentId);
  }

  return found ? currentExploitations[i] : null;
}
```

I would also recommend making your code cleaner by passing in the variable `currentExploitations`.  It makes it MUCH easier to debug/read, because you look at all the function call to see what `currentExploitations` is.  Currently, it is impossible to know where `currentExploitations` is defined.  Here's the improved version:

```js
function getParentAccount(parentId, accounts /* currentExploitations */) {
  var found = false,
      i = accounts.length;

  // the loop decrements our counter and checks that we haven't found anything
  while (!found && i--) {
    found = (accounts[i].id === parentId);
  }

  return found ? accounts[i] : null;
}

// ... wherever you call getParentAccount()
getParentAccount(id, currentExploitations);
```

Finally, if you don't like returning a ternary operator (the `?` sign), you can also just write this:

```js
function getParentAccount(parentId, accounts) {
  var found = false,
      i = accounts.length;

  // the loop decrements our counter and checks that we haven't found anything
  while (!found && i--) {
    found = (accounts[i].id === parentId) ? accounts[i].id : false;
  }

  return found;
}

// ... wherever you call getParentAccount()
getParentAccount(id, currentExploitations);
```

Personally, I prefer to do the assignment outside of the loop, but it's up to personal preference.

server/controllers/reports_proposed/data/variation_exploitation.js
```diff
@@ -0,0 +1,174 @@
...
+      break;
+    }
+  };
+  return found;
+}
+
+function getSubAccountsOnly (list){
+  return list.filter(function (item){
+    return item.account_type_id === 1;
```
jniles: Is a SubAccount an income/expense account?  Will "income/expense" always have id 1 in every database we install bhima?

server/controllers/reports_proposed/data/variation_exploitation.js
```diff
@@ -0,0 +1,174 @@
...
+      previousSubAccounts[i].percent = getPercent(previousSubAccounts[i].balance - previousSubAccounts[i].previousBalance, previousSubAccounts[i].previousBalance);
+      previousSubAccounts[i].variationValue = getNumeralFormat(previousSubAccounts[i].balance - previousSubAccounts[i].previousBalance);
+
+      founds.push(previousSubAccounts[i]);
+      pTotalBalance +=previousSubAccounts[i].previousBalance ? previousSubAccounts[i].previousBalance : 0;
+      inds_p.push(i);
+    }
+  };
+
+  inds_c.forEach(function (item){
```
jniles: It is not at all clear what you are doing here.  What are you doing?  Can you provide some comments?

server/controllers/reports_proposed/data/variation_exploitation.js
```diff
@@ -0,0 +1,174 @@
...
+  return previous;
+}
+
+function getNumeralFormat (nb){
+  return numeral(nb ? nb : 0).format(formatDollar);
+}
+
+//calculate the percent of x in y
+
+function getPercent(x, y){
```
jniles: This can be one line:
```js
function getPercent(x,y) {
  return (y === 0) ? 100 : (x / y) * 100;
}
```
It's a bit strange to always return 100 if the second argument is 0.  But I guess if it makes sense in your use case, that's alright.

server/controllers/reports_proposed/templates/variation_exploitation.dot
```diff
@@ -0,0 +1,99 @@
...
+      <th>{{=it.i18n.COMPTE}}</th>
+      <th>{{=it.i18n.DESIGNATION}}</th>
+      <th>{{=it.i18n.PREVIOUS}}</th>
+      <th>{{=it.i18n.CURRENT}}</th>
+      <th>{{=it.i18n.VARIATION_PERCENT}}</th>
+      <th>{{=it.i18n.VARIATION_MONTANT}}</th>
+    </thead>
+    <tbody>
+      {{~it.data : account : ind}}
+        {{? account.account_type_id == 3 && account.is_charge == 1}}
```
jniles: We probably shouldn't hard code database IDs into our application logic.

server/controllers/reports_proposed/data/variation_exploitation.js
```diff
@@ -0,0 +1,174 @@
...
+  context.i18n = i18VariationExploitation;
+  context.options = options;
+
+
+  var sql = 
+    'SELECT account.id, account.account_number, account.account_txt, account.parent, account.is_charge, account.account_type_id, (CASE WHEN account.is_charge = 1 THEN SUM(period_total.debit - period_total.credit) ELSE SUM(period_total.credit - period_total.debit) END) AS balance ' +
+    'FROM account LEFT JOIN period_total on account.id = period_total.account_id WHERE account.is_ohada = 1 AND account.account_type_id in (?,?) AND (period_total.fiscal_year_id in (?) OR ISNULL(period_total.fiscal_year_id)) ' +
+    'AND account.id NOT IN (SELECT account.id FROM account WHERE account.account_type_id = ?) GROUP BY account.id';
+
+  db.exec(sql, [1, 3, options.fy, 2])
```
jniles: Where did the numbers 1, 3, and 2 come from?

server/controllers/reports_proposed/data/variation_exploitation.js
```diff
@@ -0,0 +1,174 @@
...
+      ind = i;
+      break;
+    }
+  };
+  if(ind >= 0){previousSubAccounts.splice(ind, 1);}else{previous.balance = 0;}
+  return previous;
+}
+
+function getNumeralFormat (nb){
+  return numeral(nb ? nb : 0).format(formatDollar);
```
jniles: or `numeral(nb || 0).format(formatDollar);`

server/controllers/reports_proposed/data/variation_exploitation.js
```diff
@@ -0,0 +1,174 @@
...
+      previousSubAccounts[i].viewPreviousBalance = getNumeralFormat(previousSubAccounts[i].balance);
+
+      previousSubAccounts[i].balance = 0;
+      previousSubAccounts[i].viewBalance = getNumeralFormat(previousSubAccounts[i].balance);
+
+      previousSubAccounts[i].percent = getPercent(previousSubAccounts[i].balance - previousSubAccounts[i].previousBalance, previousSubAccounts[i].previousBalance);
+      previousSubAccounts[i].variationValue = getNumeralFormat(previousSubAccounts[i].balance - previousSubAccounts[i].previousBalance);
+
+      founds.push(previousSubAccounts[i]);
+      pTotalBalance +=previousSubAccounts[i].previousBalance ? previousSubAccounts[i].previousBalance : 0;
```
jniles: or `pTotalBalance += (previousSubAccounts[i].previousBalance || 0);`

server/controllers/reports_proposed/data/variation_exploitation.js
```diff
@@ -0,0 +1,174 @@
...
+  return found;
+}
+
+function getSubAccountsOnly (list){
+  return list.filter(function (item){
+    return item.account_type_id === 1;
+  });
+}
+
+function getAllChild (parentId){
```
jniles: This function seems to do a lot more than getting all children.  Would you consider renaming/commenting it so that future developers understand what is happening here?

server/controllers/reports_proposed/templates/variation_exploitation.dot
```diff
@@ -0,0 +1,99 @@
+<head>   
+  <link rel="stylesheet" href="{{=it.path}}/templates/style/normalize.css">
+  <link rel="stylesheet" href="{{=it.path}}/templates/style/skeleton.css">
+  <link rel="stylesheet" href="{{=it.path}}/templates/style/reports.css">  
+</head>
+<meta charset="utf-8"/>
```
jniles: This should be in the `<head>`.

client/src/partials/reports_proposed/variation_exploitation/variation_exploitation.js
```diff
@@ -0,0 +1,69 @@
...
+
+  validate.process(dependencies)
+    .then(setDefaultConfiguration);
+
+  function setDefaultConfiguration(models) {
+    angular.extend(vm, models);
+    selectConfiguration('language', configuration.language.options[1]);
+    vm.fiscal_year_id = vm.fiscalYears.data[vm.fiscalYears.data.length-1].id;
```
jniles: What happens if there are no fiscal years?

client/src/partials/reports_proposed/variation_exploitation/variation_exploitation.html
```diff
@@ -0,0 +1,46 @@
...
+    <div class="panel-heading">{{'VARIATION_EXPLOITATION.DOCUMENT_CONFIG' | translate}}</div>
+      <div class="panel-body">
+        <form>
+
+          <div class="form-group">
+            <label class="control-label col-xs-4 required">
+              {{ "VARIATION_EXPLOITATION.FISCAL_YEAR" | translate }}
+            </label>
+            <div>
+              <select class="form-bhima" ng-model="variationCtrl.fiscal_year_id" ng-options="fy.id as fy.fiscal_year_txt for fy in variationCtrl.fiscalYears.data" required>
```
jniles: You probably should provide a default `<option>` element for the user as a placeholder.  Selecting in the controller as you did [here](https://github.com/IMA-WorldHealth/bhima/pull/837/files#diff-516a7d58993316077d75d5d7e37c182fR24) will fail if there are no fiscal years.

client/src/partials/reports_proposed/variation_exploitation/variation_exploitation.html
```diff
@@ -0,0 +1,46 @@
...
+  <div class="panel panel-default">
+    <div class="panel-heading">{{'VARIATION_EXPLOITATION.DOCUMENT_CONFIG' | translate}}</div>
+      <div class="panel-body">
+        <form>
+
+          <div class="form-group">
+            <label class="control-label col-xs-4 required">
+              {{ "VARIATION_EXPLOITATION.FISCAL_YEAR" | translate }}
+            </label>
+            <div>
```
jniles: You don't need this <div>.

client/src/partials/reports_proposed/variation_exploitation/variation_exploitation.html
```diff
@@ -0,0 +1,46 @@
...
+</header>
+
+<main>
+  <div class="panel panel-default">
+    <div class="panel-heading">{{'VARIATION_EXPLOITATION.DOCUMENT_CONFIG' | translate}}</div>
+      <div class="panel-body">
+        <form>
+
+          <div class="form-group">
+            <label class="control-label col-xs-4 required">
```
jniles: You don't need the classes `control-label` or `col-xs-4`.  Those are only useful when you are creating a Horizontal or Inline form.

Also, the `col-xs-4` class makes the label not line up with the `<select>` or `<input>` element.  Here are two screenshots from reports, one that is a copy and paste of this code and another that was written without the `control-label` and `col-xs-4`.

#### Has col-xs tag
col-xs-report

#### No col-xs tag
no-col-xs-report

While it won't break the view, you can tell that the one missing the col-xs tag lines up much better with the input.

client/src/partials/reports_proposed/variation_exploitation/variation_exploitation.js
DedrickEnc: I like to expose comtroller function to the scope at the bottom and initialize some value at the top but I will put them at the bottom.

server/controllers/reports_proposed/data/variation_exploitation.js
DedrickEnc: Actually, I need it see the line 35.

server/controllers/reports_proposed/data/variation_exploitation.js
DedrickEnc: The ```currentExploitations``` scope is global but no matter

DedrickEnc:
```function getParentAccount(parentId, accounts) {
  var found = false,
      i = accounts.length;
  // the loop decrements our counter and checks that we haven't found anything
  while (!found && i--) {
    found = (accounts[i].id === parentId) ? accounts[i].id : false;
  }

  return found;
}
```

This code is correct about the syntax but not semantically. The function must return back an account or null value if there is no account found instead of returning ```false``` eventually.
I am okay with the first proposal.

server/controllers/reports_proposed/data/variation_exploitation.js
DedrickEnc: I know you want to talk about hard coded value, since now the id still 1. I will change it

client/src/partials/reports_proposed/variation_exploitation/variation_exploitation.js
DedrickEnc: I am making it required

DedrickEnc:
@jniles you can review

client/src/partials/reports_proposed/variation_exploitation/variation_exploitation.js
jniles: It would be good to have a user-friendly error message, in the case that the fiscal year doesn't exist.  Since many reports will need this validation check, I'd put it in the reportConfigService.

client/src/partials/reports_proposed/variation_exploitation/variation_exploitation.html
```diff
@@ -0,0 +1,51 @@
...
+      {{'VARIATION_EXPLOITATION.DOCUMENT_CONFIG' | translate}}
+    </div>
+    <div class="panel-body">
+      <form>
+
+        <div class="form-group">
+          <label class="required">
+            {{ "VARIATION_EXPLOITATION.FISCAL_YEAR" | translate }}
+          </label>
+          <div>
```
jniles: You still do not need this `<div>`.

client/src/partials/reports_proposed/variation_exploitation/variation_exploitation.js
jniles: Thanks.

server/controllers/reports_proposed/data/variation_exploitation.js
```diff
@@ -0,0 +1,180 @@
...
+
+  db.exec(sql, [INCOME_EXPENSE_ACCOUNT_ID, TITLE_ACCOUNT_ID, options.fy, BALANCE_ACCOUNT_ID])
+  .then(function (ce){ 
+    currentExploitations = ce || [];
+    return db.exec(sql, [INCOME_EXPENSE_ACCOUNT_ID, TITLE_ACCOUNT_ID, options.pfy, BALANCE_ACCOUNT_ID]);
+  })
+  .then(function (pe){
+    previousExploitations = pe || [];
+    context.data = process(currentExploitations, previousExploitations).sort(function (a, b){ return a.account_number - b.account_number;});
+    deferred.resolve(context);
```
jniles: Or, you can just do `return context;` and never use `q`.  That's what I mean by the comment "you don't need `q` in this module".

In general, the only time you need to make a promise using `q.defer()` is in these scenarios:
 1. You are wrapping a callback function which does not have a promise interface.  In other words, you are _creating_ your own promise interface.  Here's an example:

 ```js
var readFilePromise = function (filename, options) {
  var dfd = q.defer();
  fs.readFile(filename, options, function (error, data) {
     if (error) { return dfd.reject(error); }
     dfd.resolve(data);
  });
  return dfd.promise;
};
```
 In the above code, a promise interface didn't exist, so we needed to create one.
 2. You need to mask an API to make it look asynchronous, when it could be either asynchronous or synchronous.  For example, let's create a function that sends an HTTP request if there the user provides a numeric ID, but on any other input returns an error.

 ```js
function maybeAsync(id) {
  var dfd = q.defer();
  if (typeof id !== 'number') { return dfd.reject(new Error('Sorry, must be a number')); }
  $http.get('some/route/:id'.replace(':id', id))
  .then(dfd.resolve).catch(dfd.reject).finally();
  return dfd.promise;
}
```
 In the second case, you can usually find a way to avoid creating a new promise (by using `q.reject()` for example), but sometimes you do need to create a new one (if you are conditionally sending multiple requests for example).

These are the only good reasons, as far as I'm concerned, for creating a new promise.  Promises are useful, but dangerous because they can [swallow errors](http://jamesknelson.com/are-es6-promises-swallowing-your-errors/).

server/controllers/reports_proposed/data/variation_exploitation.js
jniles: This still has not been addressed.  We should not use hard coded ids in our database.

server/controllers/reports_proposed/data/variation_exploitation.js
```diff
@@ -0,0 +1,180 @@
...
+// Collects and aggregates data for the enterprise bilan
+
+var q       = require('q');
+var db      = require('../../../lib/db');
+var numeral = require('numeral');
+
+var formatDollar = '$0,0.00', variationExploitationDate = new Date(), accountList = [],
+    subAccounts = [], finalList = [], previousExploitations = [],
+    currentExploitations = [], previousSubAccounts = [],
+    INCOME_EXPENSE_ACCOUNT_ID = 1, TITLE_ACCOUNT_ID = 3, BALANCE_ACCOUNT_ID = 2 ;
```
jniles: These should probably be looked up from the database.

server/controllers/reports_proposed/templates/variation_exploitation.dot
```diff
@@ -0,0 +1,99 @@
...
+      <th>{{=it.i18n.COMPTE}}</th>
+      <th>{{=it.i18n.DESIGNATION}}</th>
+      <th>{{=it.i18n.PREVIOUS}}</th>
+      <th>{{=it.i18n.CURRENT}}</th>
+      <th>{{=it.i18n.VARIATION_PERCENT}}</th>
+      <th>{{=it.i18n.VARIATION_MONTANT}}</th>
+    </thead>
+    <tbody>
+      {{~it.data : account : ind}}
+        {{? account.account_type_id == 3 && account.is_charge == 0}}
```
jniles: We shouldn't rely on hard coded ids from our database.

jniles:
@dedrickenc, I made some more comments to address.

server/controllers/reports_proposed/data/variation_exploitation.js
jniles: Oh, I found an excellent resource that describes this issue.  It's called the [Deferred Anti-Pattern](https://github.com/petkaantonov/bluebird/wiki/Promise-anti-patterns#the-deferred-anti-pattern).  An "anti-pattern" is a way that people write code that is either not recommended or wrong.

PR #843, opened 2015-10-26
Fixing patient distribution and fiscal year process

DedrickEnc:
In this PR the patient distribution is fixed, there was an error introduced naively and that is bad.
I also noticed that the fiscal year code was changed to use aliasing but even there the change was not right.
I forced user to post all fiscal year data to the general ledger before closing the fiscal year, I also removed the period zero

### DIFF
--- a/server/controllers/fiscal.js
+++ b/server/controllers/fiscal.js
@@ -103,18 +103,22 @@ function createOpeningBalances(data) {
       balances = data.balances,
       totals;
 
+  // sql =
+  //   'SELECT id FROM period WHERE period_number = 0 AND fiscal_year_id = ?;';
+
   sql =
-    'SELECT id FROM period WHERE period_number = 0 AND fiscal_year_id = ?;';
+    'SELECT id FROM period WHERE fiscal_year_id = ? and period_start = ' +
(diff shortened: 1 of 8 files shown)


server/controllers/fiscal.js
```diff
@@ -176,7 +180,7 @@ function createPeriods(fiscalYearId, start, end) {
     'INSERT INTO period (fiscal_year_id, period_number, period_start, period_stop) VALUES ';
 
   // create the zero period
-  template.push([fiscalYearId, 0, periodStart, periodStart]);
+  // template.push([fiscalYearId, 0, periodStart, periodStart]);//after discussion, we will not use a period zero
```
jniles: You can just delete this line.

server/controllers/fiscal.js
```diff
@@ -103,18 +103,22 @@ function createOpeningBalances(data) {
       balances = data.balances,
       totals;
 
+  // sql =
```
jniles: You can just delete this line.

jniles:
In the future, this should be two separate pull requests.  But it looks good, thanks!

PR #846, opened 2015-10-26
Minors fixes in purchase order modules

DedrickEnc:
In this PR, I removed the ng-strict-di directive temporarily and fix some issue about date and so one ...


jniles:
Didn't we discover that `ng-strict-di` wasn't the problem?  In fact, `ng-strict-di` pointed us to the solution. Why are you removing it?

DedrickEnc:
Actually no, there are some module which break because of ng-strict-di.
I removed that temporarily, it can be added later.

DedrickEnc:
@jniles this PR can be merged any time.

jniles:
Looks good to me.  Thanks!

Issue #851, opened 2015-10-27
Employee Salary Advance Broken


jniles:
The salary advance module (`/primary_cash/expense/payday_advance/1`) seems to be broken.  It seems to suffer from a MySQL parsing error.  Shouldn't be hard to fix.

Issue #850, opened 2015-10-27
Salary Payment and Partial Salary Payment broken


jniles:
The modules salary payment (`/primary_cash/expense/salary_payment/:id`) and partial salary payment (`/primary_cash/expense/partial_payment/:id`) are broken, generating errors from `journal.js`.  They seem to be MySQL parsing errors.

PR #852, opened 2015-10-27
FIX(user_session): remove refs to user_session

### LINKED ISSUE #844: Remove insecure user_session route
In several places in bhima, we use the following code to download the user details
```js
   dependencies.user = {
     query : '/user_session'
   };
```
This is poor practice for many reasons: it is insecure, brittle, and we have the user session available immediately with the `SessionService` without sending an extra HTTP request.  We should remove instances of the above code in bhima.

jniles:
This commit fixes #844.  The ancient user_session route predated even
appstate and is highly insecure.  This commit removes all client-side
references to the /user_session route, replacing them with
SessionService's user object.

**Note**: I've tested all module except for:
 1. payday_advance
 2. partial_payment
 3. salary_payment
 since these modules are broken (see #851 and #850).  Any errors found though, are super, super easy to fix.


DedrickEnc:
I like

PR #857, opened 2015-10-27
FIX(stock dashboard): proper loading indicators

jniles:
This commit adds proper behavior to loading indicators on the stock
dashboard page, causing them to disappear if there is no data loaded.
We also add a message to the user, indicating that no data was found.

Finally, we've updated the SQL link to the stock management page, which
was recently broken.

### DIFF
--- a/client/src/partials/stock/dashboard/dashboard.html
+++ b/client/src/partials/stock/dashboard/dashboard.html
@@ -28,8 +28,15 @@
             </tr>
           </thead>
           <tbody>
-            <tr>
-              <td colspan="3" ng-if="StockDashCtrl.consumption.length === 0"><loading-indicator></loading-indicator></td>
+            <tr ng-if="StockDashCtrl.indicators.consumption">
+              <td colspan="3" class="text-center">
+                <loading-indicator></loading-indicator>
+              </td>
+            </tr>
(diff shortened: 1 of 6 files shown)


client/src/partials/stock/dashboard/dashboard.html
```diff
@@ -141,8 +148,15 @@
             </tr>
           </thead>
           <tbody>
-            <tr>
-              <td colspan="4" ng-if="StockDashCtrl.donations.length === 0"><loading-indicator></loading-indicator></td>
+            <tr ng-if="StockDashCtrl.indicators.donations">
```
DedrickEnc: Just curious, how an object called ```indicators``` can have a property ```donations```.
Simple variable should not be in plural, except array.

client/src/partials/stock/dashboard/dashboard.js
```diff
@@ -12,6 +12,12 @@ function StockDashboardController($q, $http, DataService, Dates) {
   vm.purchaseorders = {};
   vm.limits = { consumption : 10, donations : 10 };
 
+  // loading indicators
+  vm.indicators = {
```
DedrickEnc: Simple variable should not be in plural, except array but it is not a big problem

jniles: Why not?

PR #859, opened 2015-10-27
CHORE(depot): migrate stock links to depots

jniles:
This commit completes the migration of application `/stock` url links to
the `depots/:uuid` endpoint.  It also fixes an SQL error in the depot
losses requests.


client/src/partials/depots/entry/entry.js
```diff
@@ -0,0 +1,334 @@
...
+    this.inventory_uuid = null;
+    this.purchase_order_uuid = null;
+    this.expiration_date = new Date();
+    this.date = new Date();
+    this.lot_number = null;
+    this.tracking_number = uuid();
+    this.quantity = 0;
+  }
+
+  $scope.addLot = function addLot (drug) {
```
DedrickEnc: As you are using aliasing you should expose those function the the ```vm```.

jniles: Because it would take to long to rewrite the module and test that it all works.  I'm just trying to change the links in this PR.

PR #864, opened 2015-10-27
Amending account and journal voucher functionnality

DedrickEnc:
In this PR, I restricted the user to don't use the future date to pass a transaction and remove the dependence to account_number in account


jniles:
LGTM.

PR #865, opened 2015-10-27
FIX(receipts): load enterprise location properly

### LINKED ISSUE #831: Missing enterprise data in receipts
The majority of our receipts no longer load enterprise information correctly.  Here's a screenshot from the distribution to services receipt.

missingenterprisedata

Note that where it said "205" is a hard-coded value, which should have been replaced with real data (see #694).

In many ways, receipts are some of the most important pieces of the application, since the patients **must** have confidence in them to have confidence in the enterprise.  Someone should do a full audit of all receipts to make sure they are loading all the data they need, and add any data that is missing before next install.

As a side note, it would be nice to start using the `<address>` HTML element for the enterprise location.  It makes it much more semantic and pleasing to read (for humans).  Not sure it's a priority, but if this requires HTML changes, might as well fix it while you are there.

jniles:
Fixes #831.

This commit fixes a long standing bug introduced by bhima's session
service, which removed the property location_id from the project
appstate object. This has been corrected.

This commit also delete receipts_unused, which was, as the name implies,
unused.


jniles:
@dedrickenc, this PR is critical to merge in before Tshikaji.  It fixes all the location in our receipts.  See below, compared to #831.

receiptlocationdata

PR #874, opened 2015-10-28
FIX(primary cash): remove thumb icons in interface

jniles:
This commit removes the thumbs-up and thumbs-down glyphicons from the
primary cash interface.  They originally were a joke by @sfount, but
seemed to have persisted into production.  They have been replaced by
import/export type icons which do not have the same significance as the
previous set.


DedrickEnc:
???

PR #881, opened 2015-10-29
FIX(login): prevent permissionless login

### LINKED ISSUE #681: Shouldn't be able to login without permissions
You can currently log into bhima even without permissions if your user exists.  This is confusing because the user will not have a tree built (and potentially dangerous since they can still navigate via the URL bar).  We should prevent this by blocking users from logging in if they do not have permissions, potentially with a custom error message.

jniles:
Fixes #681.

This commit adds a check to ensure the user has permissions before login
into bhima. If the user does not have any permissions, we display an error
telling them to contact the system administrator.

**Screenshot** (FR)
permissionlesserror

### DIFF
--- a/client/src/partials/auth/login.js
+++ b/client/src/partials/auth/login.js
@@ -14,7 +14,7 @@ function LoginController($scope, $translate, $location, $http, $timeout, Appcach
 
   // contains the values from the login form
   vm.credentials = {};
-  vm.$error = false;
+  vm.error = false;
   vm.login = login;
   vm.setLanguage = setLanguage;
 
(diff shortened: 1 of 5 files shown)


client/src/partials/auth/login.js
```diff
@@ -78,7 +78,17 @@ function LoginController($scope, $translate, $location, $http, $timeout, Appcach
       $location.url('/');
     })
     .catch(function (error) {
-      vm.$error = true;
+
+      console.log('The error was:', error);
```
DedrickEnc: remove this console.log

jniles:
@dedrickenc fixed.

PR #897, opened 2015-10-30
Amending bilan pdf report

DedrickEnc:
This PR fixes bilan pdf report, but amending the structure of the report.

### DIFF
--- a/server/controllers/reports_proposed/data/bilan.js
+++ b/server/controllers/reports_proposed/data/bilan.js
@@ -186,33 +186,34 @@ exports.compile = function (options) {
     function getGroupReferences (section, list){
       var greferences = [];
 
-      for(var i = 0; i <= list.length - 1; i++){
-        if(list[i].sectionBilanId == section.sectionBilanId){
-          if(!exist(list[i], greferences, 'greferenceId')){
+      list.forEach(function (item){
+        if(item.sectionBilanId === section.sectionBilanId){
+          if(!exist(item, greferences, 'greferenceId')){
             greferences.push({
-              greferenceId : list[i].greferenceId,
-              greferenceAbbr : list[i].greferenceAbbr,
-              greferencePosition : list[i].greferencePosition,
-              greferenceLabel : list[i].greferenceLabel,
+              greferenceId : item.greferenceId,
+              greferenceAbbr : item.greferenceAbbr,
+              greferencePosition : item.greferencePosition,
+              greferenceLabel : item.greferenceLabel,
               refs : []
             });
           }
         }
-      }
+      });
+
       return greferences;
     }
 
     function getReferences (greference, list){
       var references = [];
 
-      for(var i = 0; i <= list.length - 1; i++){
-        if(list[i].greferenceId == greference.greferenceId){
-          if(!exist(list[i], references, 'referenceId')){
+      list.forEach(function (item){
+        if(item.greferenceId == greference.greferenceId){
+          if(!exist(item, references, 'referenceId')){
             references.push({
-              referenceId : list[i].referenceId,
-              referenceAbbr : list[i].referenceAbbr,
-              referencePosition : list[i].referencePosition,
-              referenceLabel : list[i].referenceLabel,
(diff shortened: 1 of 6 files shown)


server/controllers/reports_proposed/data/bilan.js
```diff
@@ -202,17 +183,17 @@ exports.compile = function (options) {
...
     }
 
-    function getGroupReferences (section, currents){
+    function getGroupReferences (section, list){
       var greferences = [];
 
-      for(var i = 0; i <= currents.length - 1; i++){
-        if(currents[i].sectionBilanId == section.sectionBilanId){
-          if(!exist(currents[i], greferences, 'greferenceId')){
+      for(var i = 0; i <= list.length - 1; i++){
```
jniles: Why not use an `array.forEach()` loop here?

server/controllers/reports_proposed/data/bilan.js
```diff
@@ -221,17 +202,17 @@ exports.compile = function (options) {
...
     }
 
-    function getReferences (greference, currents){
+    function getReferences (greference, list){
       var references = [];
 
-      for(var i = 0; i <= currents.length - 1; i++){
-        if(currents[i].greferenceId == greference.greferenceId){
-          if(!exist(currents[i], references, 'referenceId')){
+      for(var i = 0; i <= list.length - 1; i++){
```
jniles: Since this code is almost a duplicate of the one above, see the comment above.

server/controllers/reports_proposed/templates/bilan.dot
```diff
@@ -2,143 +2,129 @@
...
                   {{~}}
               {{~}}
               <tr>
                 <td></td>
-                <td><b>{{=it.total}} {{=section.sectionBilanLabel}} ({{=section.sectionBilanPosition}})</b></td>
-                <td><b>{{=section.totalBrut_view}}</b></td>
-                <td><b>{{=section.totalAmortProv_view}}</b></td>
-                <td><b>{{=section.totalNet_view}}</b></td>
-                <td><b>{{=section.totalPreviousNet_view}}</b></td>
+                <td class="text-center"><b>{{=it.i18nBilan.TOTAL}} {{=section.sectionBilanLabel}} ({{=section.sectionBilanPosition}})</b></td>
```
jniles: Did you try using a single "text-center" class on the `<tr>` element?  It might be clearer.

If you are using `</td><b></b></td>`, you probably mean `<th></th>`.  Why not use that instead?

server/controllers/reports_proposed/templates/bilan.dot
```diff
@@ -2,143 +2,129 @@
...
+                <td class="text-center"><b>{{=section.totalBrut_view}}</b></td>
+                <td class="text-center"><b>{{=section.totalAmortProv_view}}</b></td>
+                <td class="text-center"><b>{{=section.totalNet_view}}</b></td>
+                <td class="text-center"><b>{{=section.totalPreviousNet_view}}</b></td>
               </tr>
           {{~}}
           <tr>
             <td></td>
```
jniles: An empty element is very difficult to distinguish from an error.  Could you use colspan or rowspan instead?

server/controllers/reports_proposed/templates/bilan.dot
DedrickEnc: this syntax is used in side the table, because I can not use ```<th>``` which means table header.

server/controllers/reports_proposed/templates/bilan.dot
DedrickEnc: Yes but the contect will not be well centered.

server/controllers/reports_proposed/templates/bilan.dot
jniles: If you just want to make bold font then (and not indicate importance), I would use a `font-weight` property in a new class.  If it is important, use the `<th>` tag, [it is allowed in the table body](http://stackoverflow.com/questions/5509163/does-html5-ban-th-cells-from-tbody).

PR #898, opened 2015-11-02
CHORE(debtor group): performance improvements

jniles:
This commit adds slight performance improvements to the debtor group
module by replacing ng-repeat options with ng-options.  It also cleans
up the HTML and styling slightly for a better UX.


DedrickEnc:
Considered that, ```$scope``` will be removed soon, I merge this PR.

PR #900, opened 2015-11-02
Fixing bugs in Payroll module

mbayopanda:
This PR fix : 
- Payroll/Configuration : Erreur de reference ER_NO_REFERENCE_ROW_2 
- Payroll/Employee : Erreur avec les dates
- Payroll/Hollyday : Erreur de reference ER_NO_REFERENCE_ROW_2 
- Payroll/Offday : Erreur avec les dates
- Payroll/Rubriques
- Payroll/Multipayroll : Ecriture dans le journal lors d'un engagement de paiement de taxe entreprise

### DIFF
--- a/client/src/partials/config_accounting/config_accounting.js
+++ b/client/src/partials/config_accounting/config_accounting.js
@@ -1,98 +1,113 @@
 angular.module('bhima.controllers')
-.controller('config_accounting', [
-  '$scope',
-  '$translate',
-  '$http',
-  'validate',
-  'messenger',
-  'connect',
-  'appstate',
-  function ($scope, $translate, $http, validate, messenger, connect, appstate) {
-    var dependencies = {},
-        session = $scope.session = {};
+.controller('ConfigAccountingController', ConfigAccountingController);
 
-    dependencies.config_accountings = {
-      query : {
-        identifier : 'id',
-        tables : {
-          'config_accounting' : { columns : ['id', 'label', 'account_id'] },
-          'account' : { columns : ['account_number', 'account_txt'] }
-        },
-        join : ['config_accounting.account_id=account.id']
-      }
-    };
+ConfigAccountingController.$inject = [
+  '$translate', '$http', 'validate', 'messenger', 'connect', 'SessionService'
+];
 
-    // Add by Chris Lomame for filtering accounts of class 6
-    $http.get('/getAccount6/').
-      success(function(data) {
-        $scope.accounts = data;
-    });
+/**
+  * Config Account Controller
+  * This controller is responsible to manage account configuration for Payroll
+  */
+function ConfigAccountingController ($translate, $http, validate, messenger, connect, SessionService) {
+  var vm           = this,
+      dependencies = {},
+      session      = vm.session = {};
 
-    function startup (models) {
-      angular.extend($scope, models);
+  dependencies.config_accountings = {
+    query : {
+      identifier : 'id',
+      tables : {
+        'config_accounting' : { columns : ['id', 'label', 'account_id'] },
+        'account'           : { columns : ['account_number', 'account_txt'] }
+      },
+      join : ['config_accounting.account_id=account.id']
     }
+  };
 
-    appstate.register('enterprise', function (enterprise) {
-      $scope.enterprise = enterprise;
-      validate.process(dependencies)
-      .then(startup);
-    });
+  dependencies.accounts = {
+    query : '/getAccount6/'
+  };
 
-    $scope.delete = function (config_accountings) {
-      var result = confirm($translate.instant('TAXES.CONFIRM'));
-      if (result) {  
-        connect.basicDelete('config_accounting', config_accountings.id, 'id')
-        .then(function () {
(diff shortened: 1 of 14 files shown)


client/src/partials/config_accounting/config_accounting.js
```diff
@@ -1,98 +1,113 @@
...
+      session.action = '';
+      session.edit = {};
+    })
+    .catch(error);
+  }
+
+  function saveNew () {
+    var record = connect.clean(session.new);
+    connect.post('config_accounting', [record], ['id'])
+    .then(function () {
```
jniles: Very nice!  This is exactly correct.  Very good looking code.

client/src/partials/employee/employee.js
```diff
@@ -142,19 +151,22 @@ function EmployeeController($scope, $translate, validate, uuid, messenger, conne
...
+    return connect.post('debitor', [debitor], ['uuid']);
   }
 
   function writeEmployee(creditor_uuid, debitor_uuid) {
     session.employee.locked = (session.employee.locked)? 1 : 0;
     session.employee.creditor_uuid = creditor_uuid;
     session.employee.debitor_uuid = debitor_uuid;
-    session.employee.dob = util.sqlDate(session.employee.dob);
-    session.employee.date_embauche = util.sqlDate(session.employee.date_embauche);
+    // session.employee.dob must have a date object
```
jniles: Good comments.

client/src/partials/hollydays/hollydays.html
```diff
@@ -117,8 +121,11 @@
           <div class="form-group">
             <label class="control-label col-xs-3 required">{{ "COLUMNS.EMPLOYEE_NAME" | translate }}</label>
             <div class="col-xs-8">
-              <select class="form-bhima" required ng-model="session.edit.employee_id" ng-options="employee.id as employee.name + '  ' + employee.postnom + ' ' + employee.prenom for employee in employees.data">
-                <option></option>
+              <select class="form-bhima"
+                ng-model="HolidayCtrl.session.edit.employee_id"
+                ng-options="employee.id as employee.name + '  ' + employee.postnom + ' ' + employee.prenom for employee in HolidayCtrl.employees.data"
+                disabled required>
```
jniles: This has both a disabled and a required tag...  Why?

client/src/partials/rubric/rubriques_payroll/rubriques_payroll.js
```diff
@@ -1,153 +1,157 @@
...
+    session.new = {};
+  }
 
-      var rubrics = $scope.rubrics.data,
+  function editSave () {
+    var rubrics = vm.rubrics.data,
         advance_exist = 0;
 
-      rubrics.forEach(function (rub) {
+    rubrics.forEach(function (rub) {
```
jniles: You can also write:
```js
var advance_exist = rubrics.some(function (rub) {
  return rub.is_advance && rub.id !== session.edit.id;
});
```

client/src/partials/rubric/rubriques_payroll/rubriques_payroll.js
```diff
@@ -1,153 +1,157 @@
...
+        if (record.abbr.length <= 4) {
+          connect.put('rubric', [record], ['id'])
+          .then(function () {
+            vm.rubrics.put(record);
+            session.action = '';
+            session.edit = {};
+            messenger.success($translate.instant('RUBRIC_PAYROLL.UPDATE_SUCCES'));
+          })
+          .catch(error);
+        } else if (record.abbr.length > 4) {
```
jniles: This can just be `else`,

client/src/partials/rubric/rubriques_payroll/rubriques_payroll.js
```diff
@@ -1,153 +1,157 @@
...
-        });
-      }  
-
-      if(session.new.is_advance && advance_exist){
-        messenger.danger($translate.instant('RUBRIC_PAYROLL.NOT_ADVANCE'));
-        session.new.is_advance = null;
-      } else if(session.new.is_advance && !advance_exist  && !session.new.is_discount){
-        messenger.danger($translate.instant('RUBRIC_PAYROLL.BE_DISCOUNT'));
-        session.new.is_advance = null;
+    if(rubrics.length){
```
jniles: You don't need to check this `if` condition.  `rubrics.forEach()` will silently do nothing if `rubrics` is an empty array (`[]`). 

Also, you could use `var advance_exist = rubrics.some(fn);` for clarity.

jniles:
This PR looks very, very good.  I've made some minor comments, but it can be merged in anytime, whether those are fixed or not.  I'll leave @DedrickEnc to merge it in.

client/src/partials/hollydays/hollydays.html
mbayopanda: For Holiday, we specify holiday data for a specific employee, if we already define an holiday for an employee we can only update label, percent day, date from or date to but not the employee id; if we have to do that, we can simply delete the holiday row and create a new holiday data for another employee...

Require because we need that employee_id

PR #901, opened 2015-11-02
FIX(balance mensuelle): Fixes report summation bug

jniles:
This commit restructures the balance_mensuelle report to pull data from
only the period totals table, rather than the combined posting_journal
and general_ledger.  This change ensures that we are dealing with only
general_ledger data.

### DIFF
--- a/client/src/partials/reports/balance_mensuelle/balance_mensuelle.html
+++ b/client/src/partials/reports/balance_mensuelle/balance_mensuelle.html
@@ -3,43 +3,41 @@
     <main class="extend">
       <div class="panel panel-default">
         <div class="panel-heading">
-          <b> {{ "REPORT.CONFIGURE_REPORT" | translate }} : </b> {{ "BALANCE_MENSUELLE.TITLE" | translate }}
+          <strong>{{ "REPORT.CONFIGURE_REPORT" | translate }}:</strong> {{ "BALANCE_MENSUELLE.TITLE" | translate }}
         </div>
 
         <form class="panel-body" name="ConfigForm" novalidate>
 
-          <div class="form-group">
-            <label class="required">
-              {{ "BALANCE_MENSUELLE.CHOOSE_PERIOD" | translate }}
-            </label>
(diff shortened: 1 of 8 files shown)


client/src/partials/reports/balance_mensuelle/balance_mensuelle.html
```diff
@@ -3,43 +3,41 @@
     <main class="extend">
       <div class="panel panel-default">
         <div class="panel-heading">
-          <b> {{ "REPORT.CONFIGURE_REPORT" | translate }} : </b> {{ "BALANCE_MENSUELLE.TITLE" | translate }}
+          <strong>{{ "REPORT.CONFIGURE_REPORT" | translate }}:</strong> {{ "BALANCE_MENSUELLE.TITLE" | translate }}
```
sfount: This could be much better represented by heading tag (h1...6) but this idea can be addressed in the bootstrap upgrade.

client/src/partials/reports/balance_mensuelle/balance_mensuelle.js
```diff
@@ -90,29 +92,98 @@ function ReportBalanceMensuelleController ($translate, $window, validate, util,
...
+        account = accounts[id];
+        account.endDebit = row.debit;
+        account.endCredit = row.credit;
+        account.account_txt = row.account_txt;
+      });
+
+      // calculate totals
+      totals = Object.keys(accounts).reduce(function (totals, key) {
+        var account = accounts[key];
+        totals.beginDebit += (account.beginDebit || 0);
```
sfount: There is a lot repetitive code here, this isn't necessarily a bad thing because it is very clear what the code does. 

This could be done by iterating through the objects keys and incrementing (as the two objects always share the same key pairs). It could also be done by modelling the object as a class and writing an increment function that accepts another class (if this is used in enough places to warrant this). 

Defining the initial object passed to the filter would help provide a semantic name for the second parameter, as well as provide a clear origin of the object keys.

PR #902, opened 2015-11-02
Fixing the debitor group state

DedrickEnc:
In this PR, I fix the pdf report of debitor group state to print the header.


jniles:
LGTM.

PR #904, opened 2015-11-03
CHORE(travis): lint repository during build

jniles:
This commit adds linting of the client and server to the Travis CI
tests for an acceptable PR.  The server currently does not pass lint
checks and these will be addressed in coming pull requests.

### DIFF
--- a/sh/travis.sh
+++ b/sh/travis.sh
@@ -1,0 +1,14 @@
+#!/bin/bash
+
+# change to the top-level bhima directory
+#cd ..
+
+# build the application
+gulp build
+
+# lint the client
+gulp lint
+
+# lint the server
+jshint server/{controllers,lib,config}/*.js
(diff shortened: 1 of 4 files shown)


sh/travis.sh
```diff
@@ -0,0 +1,14 @@
...
+#cd ..
+
+# build the application
+gulp build
+
+# lint the client
+gulp lint
+
+# lint the server
+jshint server/{controllers,lib,config}/*.js
```
sfount: @jniles I'm guessing it would be best to use gulp uniformly to lint both the client and the server? - this doesn't matter until we have addressed lint issues on the server.

jniles: Yes, I agree.  Considering that the server (potentially) going to change a fair amount, this is a quick-n-dirty way to get the job done.

PR #915, opened 2015-11-08
Fixig Stock entry report

mbayopanda:
This PR fix the dashboard and the stock entry report module


DedrickEnc:
Bruce can you resend this PR there is conflict

PR #929, opened 2015-11-19
fixing donations modules

mbayopanda:
This PR fix donations modules and reports

### DIFF
--- a/client/src/partials/stock/donation_management/donation_management.js
+++ b/client/src/partials/stock/donation_management/donation_management.js
@@ -16,17 +16,9 @@ function DonationManagementController($scope, $q, $translate, $location, $routeP
     return messenger.error('NO_DEPOT_ID');
   }
 
-
   var cache = new Appcache('donation');
 
-  $scope.project = Session.project;
-  $scope.user = Session.user;
-  $scope.enterprise = Session.enterprise;
-
-  session.crud_or_read = 'crud';
-  session.view = $translate.instant('DONATION_MANAGEMENT.SEE_ALL');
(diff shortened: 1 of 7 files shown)


client/src/partials/stock/donation_management/donation_management.js
```diff
@@ -429,49 +422,54 @@ function DonationManagementController($scope, $q, $translate, $location, $routeP
...
+      is_donation   : 1,
+      is_direct     : 0
+    };
 
-      simulateWritePurchaseLine(purchase)
-      .then(simulateWritePurchaseItems(purchase.uuid))
-      .then(updateStockPurchaseOrder(purchase.uuid))
-      .catch(handleError);
-    }
+    simulateWritePurchaseLine(purchase)
```
jniles: This can be `return simulateWritePurchaseLine(purchase)` and you can avoid using a `defer()`.

client/src/partials/stock/donation_management/donation_management.js
```diff
@@ -320,28 +291,34 @@ function DonationManagementController($scope, $q, $translate, $location, $routeP
...
+      return connect.post('donations', donation);
+    })
+    .then(function () {
+      return connect.post('donation_item', donation_items);
+    })
+    .then(function () {
+      $location.path('/stock/donation_management/report/' + document_id);
+    })
+    .then(function () {
+      messenger.success($translate.instant('STOCK.ENTRY.WRITE_SUCCESS'));
```
jniles: You can move this up to into the previous `.then()` block.

PR #930, opened 2015-11-19
Tshikaji global transaction

mbayopanda:
This PR fix global transaction reports


client/src/partials/reports/global_transaction/global_transaction.js
```diff
@@ -117,22 +117,13 @@ function ReportGlobalTransactionController (connect, $translate, validate, util,
...
     });
 
-    if (vm.enterprise.currency_id === vm.model.c) {
-      session.somCredit = sCredit;
-      session.somDebit = sDebit;
-    } else {
-      session.somCredit = sCredit * exchange.rate(sCredit,vm.model.c,new Date());
-      session.somDebit = sDebit * exchange.rate(sDebit,vm.model.c,new Date());
-    }
+    session.somCredit = Number(exchange.convertir(sCredit, vm.enterprise.currency_id, vm.model.c, new Date())).toFixed(2);
```
jniles: This should not be converted.  The SQL powering this report selects `credit_equiv AS credit` and `debit_equiv AS debit`, which are already converted into the enterprise currency.

PR #940, opened 2015-12-14
Fix operating account

lomamech:
Changing the Display of the operating account report,displaying the result on the load column or profit and hiding the account for which the charge is equal to the benefit


jniles:
Chris, can you update this to use the latest development? I didn't realize anyone was still writing code and we completely smashed the repository. 

Please update your code and resubmit.

PR #947, opened 2015-12-17
Improve report operating account

lomamech:
- Changing the Display of the operating account report,displaying the result on the load column or profit and hiding the account for which the charge is equal to the benefit
- Imporve translate file


jniles:
LGTM.

PR #952, opened 2016-02-26
Reinit to the last version at Tshikaji

mbayopanda:
This PR contains the version 1.X of BHIMA at Tshikaji in November 2015

It's require some fixes, as : 
- Exchange rate module

The database must be generate with an external file 'db.sql'


jniles:
@DedrickEnc, you should be in charge of the pull requests to bhima-1.X.  They no longer concern the main development team.

PR #953, opened 2016-03-05
Fixing select service in return medecines module

mbayopanda:
This PR fixes the `Return distribution from service` module. 

To run correctly the app, please : 
- Don't use `STRICT_ALL_TABLES` mysql mode 
- Use the external file db.sql
- Then execute the /server/models/update/synt.sql file

NOTA: The exchange rate module doesn't have a bug when we try to define a new exchange rate when we are not in `STRICT_ALL_TABLES` mysql mode


mbayopanda:
@DedrickEnc please merge this PR as soon as possible

PR #959, opened 2016-03-19
Cashflow Reports by Cashbox

mbayopanda:
Use real transactions values and convert to cash's currency


mbayopanda:
@DedrickEnc Convention equivalents values (`credit_equiv` maybe `debit_equiv`) are not corrects. They need a fixation

PR #966, opened 2016-03-28
Fix deprecated path, and journals origins

mbayopanda:
This PR add sql scripts in `synt-thsikaji.sql` for cleanning unit by removing : 
- Account/Reports/Balance Sheet
- Account/Reports/Grand Livre
- Account/Reports/Tfr
And all associated files in `app.js` at client side.

This PR add also script for updating in the journal and general ledger all `origin_id` for primary cash operations... It fixes also the origin_id for journal voucher transaction which conerns primary cash operations


mbayopanda:
@DedrickEnc We need now to fix the journal module when it updtate a transaction it must not update the origin_id of the transaction

DedrickEnc:
You could do that in this PR, rigth?

DedrickEnc:
Bruce ce PR doit etre rebaser, et je vais savoir si il a etait bien tester avant de l'executer.
Merci

PR #970, opened 2016-03-31
Customer Debts Reports

mbayopanda:
This PR update the customer_debt.js so that it's search debtor group data only in the general ledger

This PR fix also the cashflow reports and the ui style for printing


DedrickEnc:
Bruce, ce PR a besoin de rebase, le fichier synt.sql doit etre vide.

PR #981, opened 2016-07-18
Fix improve module

lomamech:
Improve some modules, and showing data, reports 

@DedrickEnc


lomamech:
@DedrickEnc 
Review the calculation of employee Payroll

lomamech:
@DedrickEnc 
Updte the origin name for Group Invoice

